PR #1, opened 2025-02-18
feat: minimal delegation storage

zhongeric:


src/MinimalDelegation.sol
```diff
@@ -3,4 +3,20 @@ pragma solidity ^0.8.23;
 
 import {IMinimalDelegation} from "./interfaces/IMinimalDelegation.sol";
 
-abstract contract MinimalDelegation is IMinimalDelegation {}
+contract MinimalDelegation {
```
snreynolds, line 6: keep abstract so you can inherit IMinimalDelegation?

src/MinimalDelegation.sol
```diff
@@ -3,4 +3,20 @@ pragma solidity ^0.8.23;
...
+
+    struct MinimalDelegationStorage {
+        mapping(bytes32 keyHash => bytes encodedKey) keyStorage;
+    }
+
+    // keccak256(abi.encode(uint256(keccak256("Uniswap.MinimalDelegation")) - 1)) & ~bytes32(uint256(0xff));
+    bytes32 private constant MINIMAL_DELEGATION_STORAGE_LOCATION =
+        0x21f3d48e9724698d61a2dadd352c365013ee5d0f841f7fc54fb8a78301ee0c00;
+
+    function _getMinimalDelegationStorage() private pure returns (MinimalDelegationStorage storage $) {
```
snreynolds, line 17: can we put this in a library?

so like the syntax here would be keyStorage.get()? i remember sometimes this is tricky but worth a shot

snreynolds, line 17: also where did the $ notation come from is that standard? i find it hard to read lol

test/BaseTest.t.sol
```diff
@@ -0,0 +1,19 @@
...
+contract BaseTest is Test {
+    MinimalDelegation public minimalDelegation;
+    uint256 signerPrivateKey = 0xa11ce;
+    address signer = vm.addr(signerPrivateKey);
+
+    function setUp() public {
+        minimalDelegation = new MinimalDelegation();
+
+        // delegate the signer to the minimalDelegation
+        vm.etch(signer, bytes.concat(hex"ef0100", abi.encodePacked(address(minimalDelegation))));
```
snreynolds, line 16: add a `delegate(signer, implAddr)` helper func?

src/lib/KeyLib.sol
```diff
@@ -0,0 +1,28 @@
...
+
+struct Key {
+    /// @dev Unix timestamp at which the key expires (0 = never).
+    uint40 expiry;
+    /// @dev Type of key. See the {KeyType} enum.
+    KeyType keyType;
+    /// @dev Whether the key is a super admin key.
+    /// Super admin keys are allowed to call into super admin functions such as
+    /// `authorize` and `revoke` via `execute`.
+    bool isSuperAdmin;
```
snreynolds, line 19: i think we can keep isSuperAdmin for now, but let's remove this comment? I think we can just allow superAdmin keys to execute any call

snreynolds: execute any external call

src/MinimalDelegation.sol
```diff
@@ -2,5 +2,14 @@
...
 
 import {IMinimalDelegation} from "./interfaces/IMinimalDelegation.sol";
+import {Key, KeyLib} from "./lib/KeyLib.sol";
+import {MinimalDelegationStorageLib} from "./lib/MinimalDelegationStorageLib.sol";
 
-abstract contract MinimalDelegation is IMinimalDelegation {}
+contract MinimalDelegation {
+    using KeyLib for Key;
+
+    function authorize(Key memory key) external returns (bytes32 keyHash) {
```
snreynolds, line 11: add a test? if you wanna do in followup pr thats fine but we should add unit tests as we go

src/lib/KeyLib.sol
zhongeric: sg!

test/MinimalDelegation.t.sol
```diff
@@ -0,0 +1,32 @@
...
+
+        Key memory fetchedKey = minimalDelegation.getKey(keyHash);
+        assertEq(fetchedKey.expiry, 0);
+        assertEq(uint256(fetchedKey.keyType), uint256(KeyType.Secp256k1));
+        assertEq(fetchedKey.isSuperAdmin, true);
+        assertEq(fetchedKey.publicKey, abi.encodePacked(mockSecp256k1PublicKey));
+    }
+
+    function test_revoke() public {
+        bytes32 keyHash = mockSecp256k1Key.hash();
```
snreynolds, line 25: authorize before revoking? otherwise i think it already doesnt exist

zhongeric: oh good catch haha

PR #7, opened 2025-02-24
support for batched calls that do not revert on failure

snreynolds:


test/MinimalDelegation.execute.t.sol
```diff
@@ -24,9 +26,29 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler {
         tokenB.mint(address(minimalDelegation), 100e18);
     }
 
-    function test_execute_fuzz_reverts(bytes32 mode) public {
+    function test_execute_reverts() public {
```
zhongeric, line 29: nit: test name could be more descriptive for reverting modes

src/libraries/ModeDecoder.sol
```diff
@@ -8,11 +8,21 @@ library ModeDecoder {
     // 1 byte           | 1 byte    | 4 bytes       | 4 bytes       | 22 bytes
     // CALL_TYPE        | EXEC_TYPE | UNUSED        | MODE_SELECTOR | MODE_PAYLOAD
     bytes32 constant BATCHED_CALL = 0x0100000000000000000000000000000000000000000000000000000000000000;
+    bytes32 constant BATCHED_CAN_REVERT_CALL = 0x0101000000000000000000000000000000000000000000000000000000000000;
+    bytes32 constant EXTRACT_EXEC_TYPE = 0x00ff000000000000000000000000000000000000000000000000000000000000;
```
zhongeric, line 12: wild idea potentially, what are the 4 bytes after the EXEC_TYPE used for? If we wanted to do partial fills, could we use a mask of bits to allow for up to N number of inner calls to revert? Hacky but could fit the standard, depending on how flexible 5792 is. Or maybe its better for the mode payload? 

another option would be to introduce new mode + payload. Or we could put in opData

snreynolds: yeah we can definitely use the other bytes that are unused! there are 4 bytes in the mode-selector that we can just add new modes to and decode the execution data however we want

zhongeric (approved):

PR #8, opened 2025-02-25
Key management interface finish

dianakocsis:
https://github.com/Uniswap/minimal-delegation/issues/4


src/MinimalDelegation.sol
```diff
@@ -3,13 +3,16 @@ pragma solidity ^0.8.23;
...
+import {EnumerableSetLib} from "solady/utils/EnumerableSetLib.sol";
+import {IKeyManagement} from "./interfaces/IKeyManagement.sol";
 
-contract MinimalDelegation is IERC7821 {
+contract MinimalDelegation is IERC7821, IKeyManagement {
     using ModeDecoder for bytes32;
     using KeyLib for Key;
+    using EnumerableSetLib for EnumerableSetLib.Bytes32Set;
 
     /// @dev The key does not exist.
```
snreynolds, line 17: Can move errors and events inside the interface

src/MinimalDelegation.sol
```diff
@@ -32,23 +35,33 @@ contract MinimalDelegation is IERC7821 {
 
     /// @dev Authorizes the `key`.
     function authorize(Key memory key) external returns (bytes32 keyHash) {
-        keyHash = key.hash();
-        MinimalDelegationStorageLib.get().keyStorage[keyHash] = abi.encode(key);
+        keyHash = _authorize(key);
```
snreynolds, line 38: this is scary, I guess we merged in the other authorize/revoke PR without checking the msg.sender 😨 . Can you add auth checks to `authorize` and `revoke`?

test/MinimalDelegation.t.sol
```diff
@@ -17,23 +17,103 @@ contract MinimalDelegationTest is DelegationHandler {
...
+        minimalDelegation.authorize(mockSecp256k1Key);
+
+        Key memory fetchedKey = minimalDelegation.getKey(keyHash);
+        assertEq(fetchedKey.expiry, 0);
+        assertEq(uint256(fetchedKey.keyType), uint256(KeyType.Secp256k1));
+        assertEq(fetchedKey.isSuperAdmin, true);
+        assertEq(fetchedKey.publicKey, abi.encodePacked(mockSecp256k1PublicKey));
+        assertEq(minimalDelegation.keyCount(), 1);
+
+        mockSecp256k1Key =
```
snreynolds, line 42: is this just changing the expiry? if so can do:

`mockSecp256k1Key.expiry = uint40(block.timestamp + 3600);`

test/MinimalDelegation.t.sol
```diff
@@ -17,23 +17,103 @@ contract MinimalDelegationTest is DelegationHandler {
...
     }
 
     function test_revoke() public {
         // first authorize the key
         bytes32 keyHash = minimalDelegation.authorize(mockSecp256k1Key);
+        assertEq(minimalDelegation.keyCount(), 1);
 
         // then revoke the key
         minimalDelegation.revoke(keyHash);
+        vm.snapshotGasLastCall("revoke");
```
snreynolds, line 64: nice thanks for adding these

test/MinimalDelegation.t.sol
```diff
@@ -17,23 +17,103 @@ contract MinimalDelegationTest is DelegationHandler {
...
+        assertEq(minimalDelegation.keyCount(), 0);
+    }
+
+    function test_revoke_revertsWithKeyDoesNotExist() public {
+        bytes32 keyHash = mockSecp256k1Key.hash();
+        vm.expectRevert(KeyDoesNotExist.selector);
+        minimalDelegation.revoke(keyHash);
+    }
+
+    function test_keyCount() public {
```
snreynolds, line 78: add a fuzz? 

`test_fuzz_keyCount(uint8 numKeys)`

src/MinimalDelegation.sol
```diff
@@ -32,23 +35,33 @@ contract MinimalDelegation is IERC7821 {
...
+    }
+
+    /// @dev Returns the number of authorized keys.
+    function keyCount() external view returns (uint256) {
+        return MinimalDelegationStorageLib.get().keyHashes.length();
+    }
+
+    /// @dev Returns the key at the `i`-th position in the key list.
+    function keyAt(uint256 i) external view returns (Key memory key) {
+        return getKey(MinimalDelegationStorageLib.get().keyHashes.at(i));
```
zhongeric, line 55: make an internal _getKey and use it both here + the external getKey function?

src/MinimalDelegation.sol
```diff
@@ -67,4 +80,19 @@ contract MinimalDelegation is IERC7821 {
             if (!success) revert IERC7821.CallFailed();
         }
     }
+
+    function _authorize(Key memory key) private returns (bytes32 keyHash) {
+        keyHash = key.hash();
+        MinimalDelegationStorage storage minimalDelegationStorage = MinimalDelegationStorageLib.get();
+        minimalDelegationStorage.keyStorage[keyHash] = abi.encode(key);
+        minimalDelegationStorage.keyHashes.add(keyHash);
```
zhongeric, line 88: check the result from .add()? I think it returns a bool

test/utils/DelegationHandler.sol
```diff
@@ -15,6 +15,10 @@ contract DelegationHandler is Test {
     address mockSecp256k1PublicKey = makeAddr("mockSecp256k1PublicKey");
     Key public mockSecp256k1Key = Key(0, KeyType.Secp256k1, true, abi.encodePacked(mockSecp256k1PublicKey));
 
+    address mockSecp256k1PublicKey2 = makeAddr("mockSecp256k1PublicKey2");
+    Key public mockSecp256k1Key2 =
+        Key(uint40(block.timestamp + 3600), KeyType.Secp256k1, false, abi.encodePacked(mockSecp256k1PublicKey2));
```
zhongeric, line 20: nit: I worry that block.timestamp here could break tests if we do anything with vm.roll / warp in the future. Maybe just pin this to be safe?

src/MinimalDelegation.sol
dianakocsis: this would return false though if its the key already exists with a different expiry

zhongeric: i think the key hash is only over the keyType + public key, maybe we should change that

dianakocsis: would it be weird to have 2 keys authorized with only different expiries? unless they revoke the old one first

snreynolds: i think the current behavior is correct but feel free to add a comment maybe?

snreynolds (approved):

PR #11, opened 2025-02-26
feat: add optional opData mode

### LINKED ISSUE #3: opData supported execution mode
implement new execution mode to decode opData
- follow spec
- reevaluate if need stronger schema for encoding popular use case

zhongeric:
- opData mode
- opData mode + canRevert execution type
- does not support empty opData

closes #3


src/MinimalDelegation.sol
```diff
@@ -50,15 +53,26 @@ contract MinimalDelegation is IERC7821 {
     }
 
     function supportsExecutionMode(bytes32 mode) external pure override returns (bool result) {
-        return mode.isBatchedCall();
+        return mode.isBatchedCall() || mode.supportsOpData();
```
zhongeric, line 56: i dont love this, doesn't seem like it will scale well

src/MinimalDelegation.sol
```diff
@@ -50,15 +53,26 @@ contract MinimalDelegation is IERC7821 {
...
         if (msg.sender != address(this)) revert IERC7821.Unauthorized();
     }
 
-    // We currently only support calls initiated by the contract itself which means there are no checks needed on the target contract.
-    // In the future, other keys can make calls according to their key permissions and those checks will need to be added.
+    // Execute a batch of calls according to the mode and any optionally provided opData
+    function _execute(bytes32 mode, Calls[] memory calls, bytes memory opData) private {
+        if (opData.length == 0) {
+            _authorizeCaller();
+            return _execute(mode, calls);
```
snreynolds, line 67: i think we should revert if the mode said it supports opData but the length is 0.. its weird to support this, feels like it could be used maliciously 

also this support in the reference impl is where I found a criticial bug. it doesnt exist in our code base yet because we dont support batch of batches but imagine if we supported batch of batches where you can re-call execute, then the verification goes through this code path where it checks msg.sender but msg.sender is already set to address(this) because it re-entered

src/MinimalDelegation.sol
snreynolds: yaa i agree but i think we basically have to do strict equality checks because i think its actually incorrect to proceed with a mode that has dirtied but unused bits

src/MinimalDelegation.sol
zhongeric: Hmm the spec says:
>     /// - If `opData` is empty, the implementation SHOULD require that `msg.sender == address(this)`.

I think calling with empty opData should be valid?

> i think we should revert if the mode said it supports opData but the length is 0.. its weird to support this, feels like it could be used maliciously also this support in the reference impl is where I found a criticial bug. it doesnt exist in our code base yet because we dont support batch of batches but imagine if we supported batch of batches where you can re-call execute, then the verification goes through this code path where it checks msg.sender but msg.sender is already set to address(this) because it re-entered

Agree on this. To me there are two classes of modes:
- Batched call (mode = 0x0000000), unrestricted, msg.sender must equal address(this)
- OpData call (mode = 0x7801000), must go through signature authorization.

If you have a call using opData, presumably it includes a signature to authorize the outermost call, and so thus you should be able to nest sensitive batched calls within it? Since the user signed off on it. It could also easily be a batch of opData calls, each with their own signature but that seems unrealistic...

snreynolds (approved):

PR #22, opened 2025-03-04
feat: refactor Calls to Call and add Library

zhongeric:
- tests will follow up when signature verification flow is built out


snreynolds (approved):
nicee just tests?

zhongeric:
Yea I'll follow up once we implement it in the delegation contract

PR #23, opened 2025-03-04
fix: add basic unit tests for CallLib

zhongeric:


test/libraries/CallLib.t.sol
```diff
@@ -0,0 +1,35 @@
...
+    function test_hash_multiple_fuzz(Call[] memory calls) public pure {
+        bytes32 actualHash = CallLib.hash(calls);
+
+        bytes memory packedHashes = new bytes(32 * calls.length);
+        for (uint256 i = 0; i < calls.length; i++) {
+            bytes32 callHash = CallLib.hash(calls[i]);
+            assembly {
+                mstore(add(add(packedHashes, 0x20), mul(i, 0x20)), callHash)
+            }
+        }
```
snreynolds, line 31: would do just a vanilla abi.encodePacked() so its comparing the assembly to something non-assembly

snreynolds (commented):
lg2m 1 comment

snreynolds (approved):

PR #24, opened 2025-03-04
feat: add Executor contract and CalldataLib

zhongeric:


src/Executor.sol
```diff
@@ -0,0 +1,58 @@
...
+abstract contract Executor {
+    using ModeDecoder for bytes32;
+
+    error InvalidTarget();
+    error InvalidKeyHash();
+
+    bytes32 public constant EOA_KEYHASH = bytes32(0);
+
+    /// @notice Check if a call can be executed by a key hash
+    function canExecute(Call memory call, bytes32 keyHash) public view returns (bool) {
```
snreynolds, line 18: add to interface? should be part of IKeyManagement ?

src/Executor.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    // Execute a single call
+    function _execute(Call memory _call, bytes32 keyHash) private returns (bool success, bytes memory output) {
+        if (!canExecute(_call, keyHash)) revert IERC7821.Unauthorized();
+        return _execute(_call);
+    }
+
+    // Low level call
+    function _execute(Call memory _call) private returns (bool success, bytes memory output) {
```
snreynolds, line 54: It seems like some renaming might be in order? 3 sep internal functions called execute, and imo only this last function is actually executing anything?

src/Executor.sol
```diff
@@ -0,0 +1,58 @@
...
+        bool shouldRevert = mode.shouldRevert();
+        for (uint256 i = 0; i < calls.length; i++) {
+            (bool success, bytes memory output) = _execute(calls[i], keyHash);
+            // Reverts with the first call that is unsuccessful if the EXEC_TYPE is set to force a revert.
+            if (!success && shouldRevert) revert IERC7821.CallFailed(output);
+        }
+    }
+
+    // Execute a single call
+    function _execute(Call memory _call, bytes32 keyHash) private returns (bool success, bytes memory output) {
```
snreynolds, line 48: and seems like this function could just be inlined in the one above?

src/Executor.sol
```diff
@@ -0,0 +1,58 @@
...
+    using ModeDecoder for bytes32;
+
+    error InvalidTarget();
+    error InvalidKeyHash();
+
+    bytes32 public constant EOA_KEYHASH = bytes32(0);
+
+    /// @notice Check if a call can be executed by a key hash
+    function canExecute(Call memory call, bytes32 keyHash) public view returns (bool) {
+        if (keyHash == EOA_KEYHASH) return true;
```
snreynolds, line 19: will canExecute even be called with keyHash == 0?

src/MinimalDelegation.sol
```diff
@@ -68,7 +72,12 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
         return _getKey(keyHash);
     }
 
+    /// @inheritdoc IERC7821
     function supportsExecutionMode(bytes32 mode) external pure override returns (bool result) {
+        return _supportsExecutionMode(mode);
```
snreynolds, line 77: can this just be a function on the Mode library?

mode.isSupported()?

src/MinimalDelegation.sol
```diff
@@ -100,31 +109,24 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
...
-    // In the future, other keys can make calls according to their key permissions and those checks will need to be added.
-    function _execute(bytes32 mode, Call[] memory calls) private {
-        bool shouldRevert = mode.shouldRevert();
-        for (uint256 i = 0; i < calls.length; i++) {
-            (bool success, bytes memory output) = _execute(calls[i]);
-            // Reverts with the first call that is unsuccessful if the EXEC_TYPE is set to force a revert.
-            if (!success && shouldRevert) revert IERC7821.CallFailed(output);
+    function _execute(bytes32 mode, Call[] memory calls, bytes calldata opData) private {
+        // The caller must be address(this) if the mode is batchedCall or opData is empty
+        if (mode.isBatchedCall() || opData.length == 0) {
```
snreynolds, line 114: this is odd/almost scary to me.. if the opData.length == 0 it HAS to be a "batched call". if someone set the mode to BATCHED_CALL_SUPPORTS_OPDATA but passed in 0 length opData we'd hit this and continue?

src/MinimalDelegation.sol
```diff
@@ -100,31 +109,24 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
...
+            _authorizeCaller();
+            return _execute(mode, calls, bytes32(0));
```
snreynolds, line 116: i feel like we shouldnt have to even call checkCanExecute in this code path, we might need to branch the logic further/maybe do 2 different internal calls based on the mode? its just wasteful cause we know already that it will return true for this case

src/MinimalDelegation.sol
```diff
@@ -100,31 +109,24 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
...
 
-    // Execute a single call
-    function _execute(Call memory _call) private returns (bool success, bytes memory output) {
-        address to = _call.to == address(0) ? address(this) : _call.to;
-        (success, output) = to.call{value: _call.value}(_call.data);
+        // Decode the opData
+        // ex. check the nonce
+
+        // Finally ensure the signature is valid and decode the keyHash
+        (bool isValid, bytes32 keyHash) = _unwrapAndValidateSignature(calls.hash(), opData);
```
snreynolds, line 123: aha i see how we might need to edit the _unwrapAndValidateSignature piece

snreynolds (commented):
some comments, overall I feel like we should re-evaluate this like nested _execute internal function pattern.. its quite hard to figure out where each line is branching to. 

also maybe some refactoring with the mode checking is needed, but essentially how I'm seeing it be refactored in this PR is instead of checking the modes and decoding the data according to the mode, we just check if the mode is valid, & optimisically decode opData, then branch off the opData length?

Trying to understand the tradeoffs of this approach vs. the current way and my basic thinking is that we still need to do custom logic depending on the mode, so branching based off opData length isnt really "future proof", as in we will still at some point have to check mode types/branch based on modes. but keeping it this way might optimize for the EOA case. 

Gonna spend some time thinking about this.

src/Executor.sol
zhongeric: yea... I'll condense a bit per your comment below but down to rename

src/MinimalDelegation.sol
zhongeric: Yea its odd, I was hoping to catch that by ensuring to authorize the caller but I think there's def a better design

PR #25, opened 2025-03-05
initial p256 support

snreynolds:


.gitignore
```diff
@@ -12,3 +12,6 @@ docs/
 
 # Dotenv file
 .env
+
+# settings
+.vscode
```
zhongeric, line 17: remove?

src/MinimalDelegation.sol
```diff
@@ -126,22 +127,28 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
...
         }
-        // not implemented
-        revert("Not implemented");
+        // Otherwise, treat the signature as a wrapped signature, and unwrap it before validating.
+        return _unwrapAndValidate(hash, signature);
+    }
+
+    /// @dev Returns if the wrapped signature is valid.
+    /// TODO: Implement WebAuthnP256 validation.
+    function _unwrapAndValidate(bytes32 digest, bytes calldata wrappedSignature) internal view returns (bool) {
```
zhongeric, line 140: For my PR and work I think it might be better to break this up even further into an unwrap and a separate validate function, since we need to decode the wrappedSignature later to pass into _execute calls

test/ERC1271.t.sol
```diff
@@ -81,7 +81,7 @@ contract ERC1271Test is DelegationHandler {
     function test_isValidSignature_invalidSignatureLength_reverts() public {
         bytes32 hash = keccak256("test");
         bytes memory signature = new bytes(63);
-        vm.expectRevert("Not implemented");
+        vm.expectRevert();
```
zhongeric, line 84: is there a custom error from the ECDSA lib?>

test/utils/DelegationHandler.sol
```diff
@@ -13,6 +13,7 @@ contract DelegationHandler is Test {
     uint256 signerPrivateKey = 0xa11ce;
     address signer = vm.addr(signerPrivateKey);
     IMinimalDelegation public signerAccount;
+    uint256 DEFAULT_EXPIRY = 10 days;
```
zhongeric, line 16: nit: DEFAULT_KEY_EXPIRY

test/utils/TestKeyManager.sol
```diff
@@ -0,0 +1,120 @@
...
+import {Vm} from "forge-std/Vm.sol";
+
+struct TestKey {
+    uint40 expiry;
+    KeyType keyType;
+    bool isSuperAdmin;
+    bytes publicKey;
+    // saved to sign messages
+    uint256 privateKey;
+}
```
zhongeric, line 15: I wonder if it would be nicer to have 

struct TestKey {
Key key
uint256 privateKey
}

so we're working directly with the underlying Key struct

test/utils/TestKeyManager.sol
```diff
@@ -0,0 +1,120 @@
...
+                isSuperAdmin: false,
+                publicKey: abi.encodePacked(defaultAddress),
+                privateKey: DEFAULT_SECP256K1_PK
+            });
+        } else {
+            revert KeyNotSupported();
+        }
+    }
+
+    // Create a public key derived froma seed.
```
zhongeric, line 54: nit: comment

zhongeric (commented):
Nice! KeyManager is sweet

test/ERC1271.t.sol
snreynolds: no cause its reverting bc of the abi.decode call, since we can process signatures whos length is not the exactly 64 or 65

src/MinimalDelegation.sol
snreynolds: yup totally, i think we will have to refactor to accomadate for ur PR but figured I could pull in your changes into this before we merge this?

snreynolds: im just not sure exactly how I want this to look until I play around with your refactor

test/utils/TestKeyManager.sol
snreynolds: I also expose like a `toKey()` function if we want the underlying Key object

it just might be awkward to do assertions about the underlying key then bc it would look like
`assertEq(testKey.key.expiry, 0)`

zhongeric: ah yeah that makes sense, lets just make sure to update this struct if we change Key

src/MinimalDelegation.sol
zhongeric: sounds good!

snapshots/MinimalDelegationExecuteTest.json
```diff
@@ -1,6 +1,6 @@
 {
-  "execute_BATCHED_CALL_native_singleCall": "38227",
-  "execute_BATCHED_CALL_singleCall": "60635",
-  "execute_BATCHED_CALL_twoCalls": "97794",
+  "execute_BATCHED_CALL_native_singleCall": "59350",
```
zhongeric, line 2: large increase in native test?

zhongeric (approved):
lgtm!

PR #28, opened 2025-03-06
Decode opdata, add calldata decoders

snreynolds:


src/libraries/CalldataDecoder.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    /// @notice equivalent to abi.decode(data, (uint256, bytes)) in calldata
+    /// @return nonce The nonce value.
+    /// @return signature The signature to verify.
+    function decodeOpdata(bytes calldata data) internal pure returns (uint256 nonce, bytes calldata signature) {
+        assembly {
+            nonce := calldataload(data.offset)
+        }
+        signature = data.toBytes(1);
+    }
```
zhongeric, line 29: I don't really like having biz logic here in the libraries, seeing as we may want to handle decoding depending on the mode, maybe we should make this more generic for types

snreynolds: you mean like call it decodeUint256Bytes or something?

src/MinimalDelegation.sol
```diff
@@ -12,33 +12,72 @@ import {Key, KeyLib} from "./libraries/KeyLib.sol";
...
-            Call[] memory calls = abi.decode(executionData, (Call[]));
+            Call[] calldata calls = executionData.decodeCalls();
             _authorizeCaller();
-            _execute(mode, calls);
+            _dispatch(mode, calls, bytes32(0));
         } else if (mode.supportsOpData()) {
-            (Call[] memory calls, bytes memory opData) = abi.decode(executionData, (Call[], bytes));
-            _execute(mode, calls, opData);
+            // TODO: Decode in calldata
+            // executionData.decodeWithOpData();
```
zhongeric, line 34: remove todo

src/MinimalDelegation.sol
```diff
@@ -12,33 +12,72 @@ import {Key, KeyLib} from "./libraries/KeyLib.sol";
...
+            bytes32 keyHash = _authorizeOpData(mode, calls, opData);
+            _dispatch(mode, calls, keyHash);
         } else {
             revert IERC7821.UnsupportedExecutionMode();
         }
     }
 
+    /// @return keyHash The keyHash used to authorize the calls.
+    /// @dev The mode is passed to allow other modes to specify different types of opData decoding.
+    function _authorizeOpData(bytes32, Call[] calldata calls, bytes calldata opData)
```
zhongeric, line 45: nit: natspec, returns keyHash

src/MinimalDelegation.sol
```diff
@@ -99,6 +138,14 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
         return abi.decode(data, (Key));
     }
 
+    function isValidSignature(bytes32 hash, bytes calldata signature) public view override returns (bytes4 result) {
```
zhongeric, line 141: nit: inheritdoc

src/MinimalDelegation.sol
```diff
@@ -12,33 +12,72 @@ import {Key, KeyLib} from "./libraries/KeyLib.sol";
...
+    }
+
+    function _dispatch(bytes32 mode, Call[] calldata calls, bytes32 keyHash) private {
+        bool shouldRevert = mode.shouldRevert();
+        for (uint256 i = 0; i < calls.length; i++) {
+            (bool success, bytes memory output) = _execute(calls[i], keyHash);
+            // Reverts with the first call that is unsuccessful if the EXEC_TYPE is set to force a revert.
+            if (!success && shouldRevert) revert IERC7821.CallFailed(output);
+        }
+    }
```
zhongeric, line 67: wdyt about ripping out the canExecute for now, we can keep `keyHash` accessible in _authorizeOpData and return it in _unwrapAndValidate signature because we may still want to restrict ERC1271 signing for example, to only the root EOA private key. That would simplify this for now

src/libraries/CalldataDecoder.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {Call} from "./CallLib.sol";
+
+library CalldataDecoder {
+    using CalldataDecoder for bytes;
+    /// @notice equivalent to SliceOutOfBounds.selector, stored in least-significant bits
+
+    error SliceOutOfBounds();
```
zhongeric, line 10: not used?

src/libraries/CalldataDecoder.sol
zhongeric: this might be nicer if implemented in the ModeDecoder, since its possible that more than one mode will have the same opData decoding.

`function decodeOpdata(bytes32 mode, bytes calldata data)`

and then do conditional branching within the function, you have access to `isBatchedCall`, etc. too

src/libraries/CalldataDecoder.sol
```diff
@@ -0,0 +1,83 @@
...
+            calls.offset := add(lengthPtr, 0x20)
+            calls.length := calldataload(lengthPtr)
+
+            let opLengthPtr := add(data.offset, calldataload(add(data.offset, 0x20)))
+            opData.offset := add(opLengthPtr, 0x20)
+            opData.length := calldataload(opLengthPtr)
+        }
+    }
+
+    /// TODO: Import CalldataDecoder.sol, has been previously audited.
```
zhongeric, line 57: still a TODO or this is the imported code?

src/MinimalDelegation.sol
snreynolds: ok yeah I see, I removed the keyHash return from `_authorizeOpData` but its accessible within that function

we can always add the keyHash path back if we need to check things

src/libraries/CalldataDecoder.sol
snreynolds: ah yeah this was a pattern we kept to just show the custom error selector `SLICE_ERROR_SELECTOR` but I can just comment it for now

src/libraries/CalldataDecoder.sol
snreynolds: TODO still, I wanted to see if we could instead import code we've written in CalldataDecorder that we use in v4-periphery but for now I just copied the bottom function in

src/MinimalDelegation.sol
```diff
@@ -108,47 +142,34 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
...
         }
         // Otherwise, treat the signature as a wrapped signature, and unwrap it before validating.
-        return _unwrapAndValidate(hash, signature);
+        (isValid,) = _unwrapAndValidateSignature(hash, signature);
     }
 
     /// @dev Returns if the wrapped signature is valid.
     /// TODO: Implement WebAuthnP256 validation.
-    function _unwrapAndValidate(bytes32 digest, bytes calldata wrappedSignature) internal view returns (bool) {
+    function _unwrapAndValidateSignature(bytes32 digest, bytes calldata wrappedSignature)
```
zhongeric, line 157: this doesn't work if the signature is sepc256k1?

snreynolds: ah yeah I think it should actually call _isValidSignature

src/MinimalDelegation.sol
```diff
@@ -20,33 +22,56 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, IAccoun
...
         if (mode.isBatchedCall()) {
-            Call[] memory calls = abi.decode(executionData, (Call[]));
+            Call[] calldata calls = executionData.decodeCalls();
             _authorizeCaller();
-            _execute(mode, calls);
+            _dispatch(mode, calls);
         } else if (mode.supportsOpData()) {
-            (Call[] memory calls, bytes memory opData) = abi.decode(executionData, (Call[], bytes));
-            _execute(mode, calls, opData);
+            // executionData.decodeWithOpData();
```
zhongeric, line 39: nit: remove this comment?

src/MinimalDelegation.sol
```diff
@@ -20,33 +22,56 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, IAccoun
...
 
+    /// @dev The mode is passed to allow other modes to specify different types of opData decoding.
+    function _authorizeOpData(bytes32, Call[] calldata calls, bytes calldata opData) private view {
+        // TODO: Can switch on mode to handle different types of authorization, or decoding of opData.
+        // For now, we only support decoding necessary information needed to verify 1271 signatures.
+        (, bytes calldata signature) = opData.decodeUint256Bytes();
+        // TODO: Nonce validation.
+        // Check signature.
+        bool isValid;
+        /// The calls are not safe hashed with _hashTypedData, as the domain separator is sufficient to prevent against replay attacks.
```
zhongeric, line 56: missing call to _isValidSignature?

zhongeric (approved):

PR #29, opened 2025-03-10
4337

dianakocsis:


src/MinimalDelegation.sol
```diff
@@ -73,6 +80,28 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712 {
...
+
+        // https://github.com/coinbase/smart-wallet/blob/main/src/CoinbaseSmartWallet.sol#L100
+        assembly ("memory-safe") {
+            if missingAccountFunds {
+                // Ignore failure (it's EntryPoint's job to verify, not the account's).
+                pop(call(gas(), caller(), missingAccountFunds, codesize(), 0x00, codesize(), 0x00))
+            }
+        }
+
+        if (_isValidSignature(userOpHash, userOp.signature)) {
```
zhongeric, line 98: Nice! I guess we don't need to re-hash the hash here like regular 1271 since the userOp includes the current address in the sender field, should be replay safe across accounts

zhongeric (approved):
lgtm!

snapshots/MinimalDelegationTest.json
```diff
@@ -1,4 +1,6 @@
 {
-  "authorize": "210070",
-  "revoke": "58185"
+  "authorize": "210092",
+  "revoke": "58203",
+  "validateUserOp with missingAccountFunds": "32839",
```
zhongeric, line 4: nit: these test names should probably follow the underscore format?

PR #31, opened 2025-03-17
feat: add opt in functionality for entrypoint

snreynolds:


src/interfaces/IERC4337Account.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {IAccount} from "account-abstraction/interfaces/IAccount.sol";
+
+interface IERC4337Account is IAccount {
```
zhongeric, line 6: nit: natspec

src/MinimalDelegation.sol
```diff
@@ -105,30 +101,24 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, IAccoun
         return mode.isBatchedCall() || mode.supportsOpData();
     }
 
+    function updateEntryPoint(address entryPoint) external {
+        _authorizeCaller();
+        MinimalDelegationStorageLib.get().entryPoint = entryPoint;
```
zhongeric, line 106: should probably emit an event here

src/ERC4337Account.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {IERC4337Account} from "./interfaces/IERC4337Account.sol";
+
+abstract contract ERC4337Account is IERC4337Account {
+    uint256 internal constant SIG_VALIDATION_FAILED = 1;
+    uint256 internal constant SIG_VALIDATION_SUCCEEDED = 0;
```
zhongeric, line 8: nit: reverse the order of these?

src/MinimalDelegation.sol
```diff
@@ -173,6 +163,11 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, IAccoun
         revert("Not implemented");
     }
 
+    function ENTRY_POINT() public view override returns (address) {
+        MinimalDelegationStorage storage minimalDelegationStorage = MinimalDelegationStorageLib.get();
```
zhongeric, line 167: nit: could this be `MinimalDelegationStorageLib.get().entrypoint` ?

zhongeric (commented):
only nits!

src/ERC4337Account.sol
snreynolds: this is defined by the spec! "SHOULD not return early when returning SIG_VALIDATION_FAILED (1). Instead, it SHOULD complete the normal flow to enable performing a gas estimation for the validation function."  https://eips.ethereum.org/EIPS/eip-4337#smart-contract-account-interface

snapshots/MinimalDelegationTest.json
```diff
@@ -1,6 +1,6 @@
 {
   "authorize": "210092",
   "revoke": "58203",
-  "validateUserOp_missingAccountFunds": "32862",
-  "validateUserOp_no_missingAccountFunds": "4178"
+  "validateUserOp_missingAccountFunds": "61864",
```
zhongeric, line 4: double in gas?

src/MinimalDelegation.sol
```diff
@@ -17,19 +17,16 @@ import {CalldataDecoder} from "./libraries/CalldataDecoder.sol";
 import {P256} from "@openzeppelin/contracts/utils/cryptography/P256.sol";
 import {PackedUserOperation} from "account-abstraction/interfaces/PackedUserOperation.sol";
 import {IAccount} from "account-abstraction/interfaces/IAccount.sol";
+import {ERC4337Account} from "./ERC4337Account.sol";
+import {IERC4337Account} from "./interfaces/IERC4337Account.sol";
```
zhongeric, line 21: don't need this import right

src/ERC4337Account.sol
zhongeric: ah yeah i just mean the order within the code, defining success before fail

zhongeric (approved):

snapshots/MinimalDelegationTest.json
snreynolds: oh these gas estimates are messed up bc they don't use via-ir

src/MinimalDelegation.sol
snreynolds: we do cause of the inheritdoc tag

PR #35, opened 2025-03-18
Add locker

zhongeric:


src/Lock.sol
```diff
@@ -0,0 +1,37 @@
...
+        if (msg.sender != address(this)) {
+            if (Locker.isLocked()) revert ContractLocked();
+            Locker.set(msg.sender);
+            _;
+            Locker.set(address(0));
+        } else if (!Locker.isLocked()) {
+            /// @dev For 7702 accounts the msg.sender is address(this) but we still need to set the lock if unset
+            Locker.set(msg.sender);
+            _;
+            Locker.set(address(0));
```
zhongeric, line 26: This is interesting, essentially we still need to lock the contract if msg.sender == address(this) for only the top level call, since this is techinically valid under 7702

src/Lock.sol
```diff
@@ -0,0 +1,37 @@
...
+/// @notice A contract that provides a reentrancy lock for external calls
+/// @author Modified from https://github.com/Uniswap/universal-router/blob/main/contracts/base/Lock.sol
+contract Lock {
+    /// @notice Thrown when attempting to reenter a locked function from an external caller
+    error ContractLocked();
+
+    /// @notice Modifier enforcing a reentrancy lock that allows self-reentrancy
+    /// @dev If the contract is not locked, use msg.sender as the locker
+    modifier isNotLocked() {
+        // Apply a reentrancy lock for all external callers
```
snreynolds, line 16: I find this a bit hard to follow all the cases.. could something like this be a little easier?

less else/if

```
address locker = _getLocker();
        if (locker != address(0)) {
            // Only allow self-reentrancy.
            if (msg.sender != address(this)) revert ContractLocked();
            _;
        } else {
            /// @dev For 7702 accounts the msg.sender is address(this) but we still need to set the lock if unset
            Locker.set(msg.sender);
            _;
            Locker.set(address(0));
        }
```

src/Lock.sol
snreynolds: makes sense, I do wonder if we also NEED to prevent external reentrancy? 

any function that authorizes through msgSender() now would fail and any other function authorizes through signatures so IF an external contract wanted to re call into the contract and had a valid signature, what is the worst case here?

src/MinimalDelegation.sol
```diff
@@ -126,7 +133,7 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, ERC4337
     }
 
     function _authorizeCaller() private view {
```
snreynolds, line 135: can you add a dev comment saying that this function should only be called if there is an enforcing locker check , since otherwise msgSender() would not be set

test/Lock.t.sol
```diff
@@ -0,0 +1,47 @@
...
+
+    bytes constant LOCKER_IS_THIS = abi.encodeWithSelector(bytes4(keccak256("lockerIsThis()")));
+    bytes constant LOCKER_IS_ANYONE = abi.encodeWithSelector(bytes4(keccak256("lockerIsAnyone()")));
+
+    function setUp() public {
+        mockLock = new MockLock();
+        addressThis = address(mockLock);
+    }
+
+    function test_lockerIsThis_succeeds() public {
```
snreynolds, line 21: assert that _getLocker() is address(0) at the end of all these calls?

zhongeric: Hmm since we are tstoring locker we would probably have to do the assert a wrapper function around the actual call to Lock?

zhongeric: Bc otherwise it's guaranteed to be cleared after each call

src/Lock.sol
zhongeric: Yea I like this

src/Lock.sol
zhongeric: I think the main outcome should be for any function using _authorizeCaller to require the top level call be initiated by address(this), which this achieves. Do we want to allow _authorize_ via signature? Like, doing a userOp to add a passkey? Like potentially...

snreynolds:
Closing, since we do not think this will be a needed feature once we can restrict calls that target address(this)

PR #40, opened 2025-03-21
Nonces

dianakocsis:


snapshots/MinimalDelegationExecuteTest.json
```diff
@@ -1,6 +1,8 @@
 {
-  "execute_BATCHED_CALL_native_singleCall": "58712",
-  "execute_BATCHED_CALL_singleCall": "57635",
-  "execute_BATCHED_CALL_twoCalls": "92827",
+  "execute_BATCHED_CALL_SUPPORTS_OPDATA_oneCall": "61343",
+  "execute_BATCHED_CALL_SUPPORTS_OPDATA_twoCalls": "96399",
```
zhongeric, line 3: nit: single vs. multiple?

src/MinimalDelegation.sol
```diff
@@ -44,16 +45,15 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, ERC4337
...
-        // TODO: Nonce validation.
+        (uint256 nonce, bytes calldata signature) = opData.decodeUint256Bytes();
+        validateAndUpdateNonce(nonce);
         // Check signature.
-        bool isValid;
-        /// The calls are not safe hashed with _hashTypedData, as the domain separator is sufficient to prevent against replay attacks.
-        (isValid,) = _isValidSignature(calls.hash(), signature);
+        (bool isValid,) = _isValidSignature(_hashTypedData(calls.hash(nonce)), signature);
         if (!isValid) revert IERC7821.InvalidSignature();
+        // TODO: Store keyHash in storage.
```
zhongeric, line 56: what is this TODO for?

src/libraries/CallLib.sol
```diff
@@ -31,7 +31,7 @@ library CallLib {
                 }
             }
 
-            return keccak256(packedHashes);
+            return keccak256(abi.encodePacked(packedHashes, nonce));
```
zhongeric, line 34: Ah I don't think the nonce should be hashed with the calls here, but rather in MinimalDelegation where we build the digest we want to verify against. This library should be strictly for operations on Call

src/libraries/MinimalDelegationStorage.sol
```diff
@@ -6,6 +6,7 @@ import {EnumerableSetLib} from "solady/utils/EnumerableSetLib.sol";
 struct MinimalDelegationStorage {
     EnumerableSetLib.Bytes32Set keyHashes;
     mapping(bytes32 keyHash => bytes encodedKey) keyStorage;
+    mapping(uint192 key => uint256 nonce) nonceSequenceNumber;
```
zhongeric, line 9: techinically this could be uint192 => uint64 for key => seq right?

src/libraries/NonceManager.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.23;
+
+/// @title NonceManager
+library NonceManager {
```
zhongeric, line 5: is this library used?

dianakocsis: oh woops meant to delete that

src/libraries/CallLib.sol
snreynolds: yeah this would break EIP712 spec actually because it needs to match a typehash

would recommend defining an outermost struct that holds the array of calls and nonce

`Execute(Call[] calls, uint256 nonce)`

src/libraries/MinimalDelegationStorage.sol
snreynolds: yeah but you can also just define it as a 

uint256 => uint256 so that you don't have to downcast all the time 

which is what I would recommend instead of passing around uint192/uint64s all the time and just document in the interface how a uint256 nonce value is read

src/NonceManager.sol
```diff
@@ -0,0 +1,23 @@
...
+/// @title NonceManager
+/// @notice A contract that manages nonces to prevent replay attacks
+abstract contract NonceManager is INonceManager {
+    /// @inheritdoc INonceManager
+    function getNonce(uint192 key) public view returns (uint256 nonce) {
+        return MinimalDelegationStorageLib.get().nonceSequenceNumber[key] | (uint256(key) << 64);
+    }
+
+    /// @inheritdoc INonceManager
+    function validateAndUpdateNonce(uint256 nonce) public {
```
snreynolds, line 16: this should be internal, otherwise anyone could invalidate your nonce!

also maybe name change?

_useNonce ?

src/interfaces/INonceManager.sol
```diff
@@ -0,0 +1,26 @@
...
+    ///               - Upper 192 bits: the provided key
+    ///               - Lower 64 bits: the expected sequence number for this key
+    function getNonce(uint192 key) external view returns (uint256 nonce);
+
+    /// @notice Validates that the provided nonce is valid and increments the sequence number
+    /// @param nonce A 256-bit value where:
+    ///             - Upper 192 bits: the sequence key
+    ///             - Lower 64 bits: must match the expected sequence number for the key
+    /// @dev If valid, increments the sequence number for future nonce validations
+    function validateAndUpdateNonce(uint256 nonce) external;
```
snreynolds, line 25: this should not be external

snreynolds (commented):
In NonceManager.sol you also need an `invalidateNonce` function to cancel nonces

& note that the downside of this approach is that every invalidation is an sstore :/

src/libraries/CallLib.sol
zhongeric: yea agree on the outermost struct

src/MinimalDelegation.sol
dianakocsis: oh weird, i thought that was already there - don't know how that got there

src/libraries/MinimalDelegationStorage.sol
dianakocsis: okay! was following what they did in entrypoint, but will switch it to this way

src/libraries/CallLib.sol
dianakocsis: oh so u mean adding another struct that contains an array of `Calls` and a nonce? that way it doesn't break EIP712? should this still be in the Call library?

zhongeric: this would probably live in a different library since it has its own typehash, etc.

src/MinimalDelegation.sol
```diff
@@ -47,12 +47,24 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, ERC4337
         }
     }
 
+    /// @inheritdoc INonceManager
+    function getNonce(uint192 key) public view override returns (uint256 nonce) {
```
snreynolds, line 51: I would just make this getNonce(uint256 key)

src/NonceManager.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    /// @dev Increments the sequence for the key in nonce (i.e. upper 192 bits).
+    /// This invalidates the nonces for the key, up to (inclusive) `uint64(nonce)`.
+    function _invalidateNonce(uint256 nonce) internal {
+        uint192 key = uint192(nonce >> 64);
+        uint256 currentSeq = MinimalDelegationStorageLib.get().nonceSequenceNumber[key];
+        uint64 targetSeq = uint64(nonce);
+        if (targetSeq < currentSeq) revert InvalidNonce();
+        MinimalDelegationStorageLib.get().nonceSequenceNumber[key] =
+            targetSeq < type(uint64).max - 1 ? targetSeq + 1 : type(uint64).max;
```
snreynolds, line 37: hm I don't know if we necessarily want to allow invalidating an entire key space at once? In p2 we limited it to `type(uint16).max` in one call

src/libraries/MinimalDelegationStorage.sol
snreynolds: yeah it does need to include the key im just saying use the types uint256 => uint256

src/NonceManager.sol
dianakocsis: it invalidates up to the seq number of the nonce, not the entire key space

src/libraries/MinimalDelegationStorage.sol
```diff
@@ -7,7 +7,7 @@ import {EnumerableSetLib} from "solady/utils/EnumerableSetLib.sol";
 struct MinimalDelegationStorage {
     EnumerableSetLib.Bytes32Set keyHashes;
     mapping(bytes32 keyHash => bytes encodedKey) keyStorage;
-    mapping(uint192 key => uint256 nonce) nonceSequenceNumber;
+    mapping(uint256 key => uint256 seq) nonceSequenceNumber;
```
snreynolds, line 10: nit but could we just call this nonces ?

the name including "number" is slightly misleading bc the data structure is a mapping

src/NonceManager.sol
snreynolds: Yeah its just somewhat of a footgun IMO to allow invalidating an entire space of nonces. It should be allowed, but can be allowed by recalling invalidateNonce

i think it should be restricted similar to p2. type(uint16).max at a time can be invalidated

src/libraries/MinimalDelegationStorage.sol
dianakocsis: hm but i like how its the same here https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/NonceManager.sol#L11-L14

snreynolds: i still think i don't like that its "number" when you are going to use that variable locally its misleading what the data type you are working with is.

we don't need to rename in this PR since I wanna just get this in, but I still think I prefer a diff name

src/NonceManager.sol
```diff
@@ -0,0 +1,47 @@
...
+    ///             - Upper 192 bits: the sequence key
+    ///             - Lower 64 bits: the sequence number to be invalidated
+    /// @dev Can't invalidate >= 2**16 nonces per transaction.
+    function _invalidateNonce(uint256 nonce) internal {
+        uint192 key = uint192(nonce >> 64);
+        uint64 currentSeq = uint64(MinimalDelegationStorageLib.get().nonceSequenceNumber[key]);
+        uint64 targetSeq = uint64(nonce);
+        if (targetSeq < currentSeq) revert InvalidNonce();
+        // Limit the amount of nonces that can be invalidated in one transaction.
+        unchecked {
```
snreynolds, line 40: I think if you're going for the design where the user has to pass in a target nonce to invalidate up to, we can just do:

```
MinimalDelegationStorageLib.get().nonceSequenceNumber[key] = targetSeq;
```

now that you have the delta check!

dianakocsis: in this case im invalidating all the nonces up to and including the one thats passed. thats why i set the new nonce to targetSeq + 1

test/libraries/CallLib.t.sol
```diff
@@ -23,10 +26,22 @@ contract CallLibTest is Test {
...
             hashes[i] = CallLib.hash(calls[i]);
         }
+
         bytes32 expectedHash = keccak256(abi.encodePacked(hashes));
+
+        assertEq(actualHash, expectedHash);
+    }
+
+    function test_hash_with_nonce_fuzz(Call[] memory calls, uint256 nonce) public pure {
+        ExecutionData memory execute = ExecutionData({calls: calls, nonce: nonce});
```
snreynolds, line 41: I think we should probs move this to a different test file but fine for now

snreynolds (approved):

PR #46, opened 2025-03-26
add 4337 e2e tests

zhongeric:
- Bump AA repo to 0.8.0 to use new entrypoint
- Add e2e
- Skip authorize opdata for entrypoint


test/utils/UserOpBuilder.sol
```diff
@@ -0,0 +1,110 @@
...
+    uint256 constant DEFAULT_PRE_VERIFICATION_GAS = 1_000_000;
+    bytes32 constant DEFAULT_GAS_FEES = bytes32(uint256(100) << 128 | uint256(100));
+
+    function initDefault() internal pure returns (PackedUserOperation memory) {
+        return PackedUserOperation(
+            address(0), 0, "", "", DEFAULT_ACCOUNT_GAS_LIMITS, DEFAULT_PRE_VERIFICATION_GAS, DEFAULT_GAS_FEES, "", ""
+        );
+    }
+
+    /// @dev Override initcode hash set to default of 0
```
snreynolds, line 20: j curious what is the override initcode hash used for?

src/MinimalDelegation.sol
```diff
@@ -47,6 +47,12 @@ contract MinimalDelegation is IERC7821, IKeyManagement, ERC1271, EIP712, ERC4337
 
     /// @dev The mode is passed to allow other modes to specify different types of opData decoding.
     function _authorizeOpData(bytes32, Call[] calldata calls, bytes calldata opData) private view {
+        if (msg.sender == ENTRY_POINT()) {
```
snreynolds, line 50: I wonder if it would at all be worth it to instead have a specific mode dedicated to userOp handling? 

also with this piece of code we should just make sure that reentrancy on the entrypoint contract can't somehow call back into this function and execute whatever

zhongeric: hmm, yeah could potentially explore because opData will be encoded differently

zhongeric: Actually I don't think we should have a separate mode for entrypoint:
- I see the decision of wrapping a call in a userOp or not as being made strictly after the calls are encoded
- Thus by adding a specific mode we require the dapp (or wallet using 5792) to know whether or not it wants to do gas sponsorship when creating the calldata, I worry this will make the experience cumbersome

snreynolds (approved):
this has the webauthn lib addition btw!

PR #50, opened 2025-03-28
WIP: add setCanExecute and pre/post call sites

zhongeric:


src/GuardedExecutor.sol
```diff
@@ -0,0 +1,92 @@
...
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @dev Represents any key hash.
+    bytes32 public constant ANY_KEYHASH = 0x3232323232323232323232323232323232323232323232323232323232323232;
+
+    /// @dev Represents any target address.
+    address public constant ANY_TARGET = 0x3232323232323232323232323232323232323232;
+
+    /// @dev Represents any function selector.
+    bytes4 public constant ANY_FN_SEL = 0x32323232;
```
snreynolds, line 27: Is there a good use case for this? 

If integrating w Universal Router this does essentially nothing, and will be true for other contracts that are weiroll style

I'm sort of against the fn selector permissions, and i guess keeping permissioning as general as possible, or as easy to move to another module as possible

PR #57, opened 2025-04-01
try to consolidate ecdsa recover flows

snreynolds:
I don't like how in the vanilla ECDSA signature case we are encoding and decoding the public key, and creating a Root Key object to operate on buttt I love that there is only one control flow to all ECDSA signature recovery

lmk if you have suggestions!

also not sure about doing all the recovery in calldata, its maybe around a 500 gas optimization from looking at snapshots


zhongeric:
I'm fine keeping the verification in memory, makes the devx a bit nicer

zhongeric:
Incorporated SignatureUnwrapper and toRootKey in #58

snreynolds:
closing since incorporated in 58

PR #58, opened 2025-04-01
Add Hooks, clean up signature flows

zhongeric:


src/libraries/SignatureUnwrapper.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {CalldataDecoder} from "./CalldataDecoder.sol";
+
+library SignatureUnwrapper {
```
zhongeric, line 6: sry i stole this haha

src/KeyManagement.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.23;
+
+import {EnumerableSetLib} from "solady/utils/EnumerableSetLib.sol";
+import {Key, KeyLib, KeyType} from "./libraries/KeyLib.sol";
+import {MinimalDelegationStorage, MinimalDelegationStorageLib} from "./libraries/MinimalDelegationStorage.sol";
+import {IKeyManagement} from "./interfaces/IKeyManagement.sol";
+
+/// @dev A base contract for managing keys.
+abstract contract KeyManagement is IKeyManagement {
```
zhongeric, line 10: just separating out into separate file

src/libraries/SignatureUnwrapper.sol
snreynolds: hehe

snreynolds: all good, I can close my other PR then

src/libraries/HookLib.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.23;
+
+import {MinimalDelegationStorage, MinimalDelegationStorageLib} from "./MinimalDelegationStorage.sol";
+import {IHook} from "../interfaces/IHook.sol";
+
+enum HookFlags {
```
snreynolds, line 7: Interesting, we have some hooks that override full account logic but some that just do before/after logic. All of the validate signature hooks are IN PLACE of signature validation whereas the execute hooks are wrapping execution logic. 

I wonder if some of the abstraction with hooks could help distinguish between the two. 

Also interesting that a hook can be associated with varying degrees of flags. ie you could have a key that uses some of the flags of the hook but not all, AND it can be changeable. This is ultimate flexibility but I wonder if we would want to make it immutable to give guarantees about using a certain key. 

IE if a user gives auth to a key for X hook which implements ALL flags but only allows it to call to some, however a user could change the flags before a txn is submitted it doesn't give great guarantees about whats going to happen. At the same time, a user could just revoke the hook entirely... Just kinda thinking out loud about this design

src/libraries/HookLib.sol
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// HookId combines an address with flags: address << 160 | flags
+type HookId is uint256;
+
+library HookLib {
+    using MinimalDelegationStorageLib for MinimalDelegationStorage;
+
+    /// @dev Get the hook for a given keyHash and flag
+    function get(bytes32 keyHash, HookFlags flag) internal view returns (IHook) {
```
snreynolds, line 23: i think this could use some library abstraction clean up but we can do it later

ie get should just return the registered validation module and THEN we should just if that validation module has the permission to be called out to

can do it as a todo tho

src/libraries/SignatureUnwrapper.sol
```diff
@@ -0,0 +1,21 @@
...
+    using CalldataDecoder for bytes;
+
+    bytes32 public constant ROOT_KEY_HASH = bytes32(0);
+
+    /// @notice Unwraps a signature.
+    /// @dev If the signature length is 64 or 65, it is not wrapped and is returned to be verified against the root key.
+    function unwrap(bytes calldata _signature) internal pure returns (bytes32 keyHash, bytes calldata signature) {
```
snreynolds, line 13: :) yay

src/MinimalDelegation.sol
```diff
@@ -124,20 +150,21 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, KeyMana
         return MinimalDelegationStorageLib.get().entryPoint;
     }
 
-    function _isValidSignature(bytes32 _hash, bytes calldata _signature)
+    /// @notice Verifies that the key signed over the digest
+    /// Handles signatures from the root ECDSA key and wrapped signatures
```
snreynolds, line 154: nit this assumes signatures are not wrapped so this is a little misleading, think u can just remove

src/libraries/HookLib.sol
zhongeric: #63

src/libraries/HookLib.sol
zhongeric: #62

PR #59, opened 2025-04-01
Add before/after execute hooks and example GuardedExecutorHook

zhongeric:
- Add execution hook call sites
- Add GuardedExecutorHook


src/MinimalDelegation.sol
```diff
@@ -52,12 +52,18 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
         if (mode.isBatchedCall()) {
             Call[] calldata calls = executionData.decodeCalls();
             _onlyThis();
-            _dispatch(mode, calls);
+            _dispatch(mode, calls, SignatureUnwrapper.ROOT_KEY_HASH);
```
snreynolds, line 55: This feels like a constant that should live in KeyLib

src/MinimalDelegation.sol
```diff
@@ -52,12 +52,18 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
...
         } else if (mode.supportsOpData()) {
-            // executionData.decodeWithOpData();
             (Call[] calldata calls, bytes calldata opData) = executionData.decodeCallsBytes();
-            _authorizeOpData(mode, calls, opData);
-            _dispatch(mode, calls);
+            (uint256 nonce, bytes calldata wrappedSignature) = opData.decodeUint256Bytes();
```
snreynolds, line 58: Can decode just return (Call[] calldata calls, uint256 nonce, bytes calldata wrappedSignature)?

src/MinimalDelegation.sol
```diff
@@ -52,12 +52,18 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
...
+            _useNonce(nonce);
```
snreynolds, line 59: It feels weird to have the top level useNonce function here to me? Usually nonce spending happens right before or in when signature verification is happening 

could we maybe make the flow in this function:

(Call[] memory calls, nonce, signature) = exeuctionData.decodeAllData();

_handleVerifySignature(signedCalls, signature)
_dispatch(mode, calls, keyHash) 

?

src/interfaces/IExecutionHook.sol
```diff
@@ -0,0 +1,21 @@
...
+
+interface IExecutionHook {
+    /**
+     * EXECUTION HOOKS
+     */
+
+    /// @dev To indicate the entire call should revert, the function MUST revert.
+    /// @param keyHash The key hash to check against
+    /// @param to The address to call
+    /// @param data TODO: The calldata sent. For `executeUserOp` calls of validation-associated hooks, hook modules
```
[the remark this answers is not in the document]
snreynolds: Are validation associated hooks/hook modules something defined by 4337?

src/interfaces/IHook.sol
```diff
@@ -1,25 +1,7 @@
...
-        view
-        returns (bytes4, bytes4);
-
-    /// @notice Validates a signature over a digest and returns a boolean
-    function overrideVerifySignature(bytes32 keyHash, bytes32 data, bytes calldata signature)
-        external
-        view
-        returns (bytes4, bool);
-}
+interface IHook is IValidationHook, IExecutionHook {}
```
snreynolds, line 7: nice i love this separation

src/hooks/shared/AccountKeyHashLib.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.23;
+
+type AccountKeyHash is bytes32;
+
+library AccountKeyHashLib {
+    /// @notice Rehash a key hash with the sender's account address
+    function wrap(bytes32 keyHash) internal view returns (AccountKeyHash) {
+        return AccountKeyHash.wrap(keccak256(abi.encode(msg.sender, keyHash)));
```
snreynolds, line 9: I think to make this more extensible it would be best to add the address as an input, then we can see from the base contract what this is wrapped with ?

src/hooks/validation/MultiSignerValidatorHook.sol
```diff
@@ -31,7 +31,7 @@ contract MultiSignerValidatorHook is IHook {
         bytes32 signerKeyHash = signerKey.hash();
 
         keyStorage[signerKeyHash] = encodedKey;
-        requiredSigners[_accountKeyHash(keyHash)].add(signerKeyHash);
+        requiredSigners[keyHash.wrap()].add(signerKeyHash);
```
snreynolds, line 34: this is interesting, so the added signer HAS to be an secpk1 key in this impl

src/MinimalDelegation.sol
```diff
@@ -69,52 +75,42 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
...
-    function _execute(Call calldata _call) private returns (bool success, bytes memory output) {
+    /// @dev Executes a low level call using execution hooks if set
+    function _execute(Call calldata _call, bytes32 keyHash) internal returns (bool success, bytes memory output) {
+        // Per ERC7821, replace address(0) with address(this)
         address to = _call.to == address(0) ? address(this) : _call.to;
+
+        IHook hook = keySettings[keyHash].hook();
+        bytes memory beforeExecuteData;
+        if(hook.hasPermission(HooksLib.BEFORE_EXECUTE_FLAG)) {
+            beforeExecuteData = hook.handleBeforeExecute(keyHash, to, _call.data);
```
snreynolds, line 108: pass call.value

src/MinimalDelegation.sol
```diff
@@ -69,52 +75,42 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
...
         }
     }
 
-    // Execute a single call.
-    function _execute(Call calldata _call) private returns (bool success, bytes memory output) {
+    /// @dev Executes a low level call using execution hooks if set
+    function _execute(Call calldata _call, bytes32 keyHash) internal returns (bool success, bytes memory output) {
+        // Per ERC7821, replace address(0) with address(this)
         address to = _call.to == address(0) ? address(this) : _call.to;
+
```
snreynolds, line 104: We still need to check the admin settings on the key here, maybe a todo comment?

src/interfaces/IExecutionHook.sol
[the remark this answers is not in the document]
zhongeric: It's actually from 6900: https://eips.ethereum.org/EIPS/eip-6900#ierc6900executionhookmodulesol

src/hooks/validation/MultiSignerValidatorHook.sol
zhongeric: yeah just for brevity for now

test/MinimalDelegation.4337.t.sol
```diff
@@ -92,14 +93,17 @@ contract MinimalDelegation4337Test is DelegationHandler, TokenHandler, ExecuteHa
...
 
         vm.prank(address(signerAccount));
         signerAccount.register(p256Key.toKey());
 
         Call[] memory calls = CallBuilder.init();
         calls = calls.push(buildTransferCall(address(tokenA), address(receiver), 1e18));
 
-        bytes memory opData = bytes("");
+        /// Pass keyHash in opData
+        /// This can be spoofed! Set it to 0 to bypass the permissions check set above
```
snreynolds, line 105: Is this just to comment that even though it can be "spoofed" it won't actually get used in the contract?

test/MinimalDelegation.4337.t.sol
```diff
@@ -92,14 +93,17 @@ contract MinimalDelegation4337Test is DelegationHandler, TokenHandler, ExecuteHa
...
+        bytes memory opData = abi.encode(bytes32(0));
         bytes memory executionData = abi.encode(calls, opData);
```
snreynolds, line 107: I feel like actually this should revert in the contract because its encoding execution data that is unexpected in the executeUserOp flow

snreynolds: it means that our CalldataDecoder lib is not handling length checks, which actually I did know about & is a TODO, but if we were to decode in memory this would actually cause a revert

src/MinimalDelegation.sol
zhongeric: good callout

test/MinimalDelegation.4337.t.sol
zhongeric: oh these tests need to be updated after adding executeUserOp

src/hooks/shared/AccountKeyHashLib.sol
zhongeric: yes good idea!

src/MinimalDelegation.sol
zhongeric: updated! it's still a bit odd because we need to fully unravel opData to get keyHash

src/MinimalDelegation.sol
zhongeric: Hmm i don't think those are equivalent

src/MinimalDelegation.sol
```diff
@@ -169,6 +165,25 @@ contract MinimalDelegation is
         else return SIG_VALIDATION_FAILED;
     }
 
+    /// @dev This function is used to handle the verification of signatures sent through execute()
+    function _handleVerifySignature(bytes32 digest, uint256 nonce, bytes memory wrappedSignature) private {
+        (bytes32 keyHash, bytes memory signature) = abi.decode(wrappedSignature, (bytes32, bytes));
```
snreynolds, line 170: can you just pass in keyHash and signature since you already unwrap above?

foundry.toml
```diff
@@ -3,7 +3,7 @@ src = "src"
 out = "out"
 libs = ["lib"]
 optimizer = true
-optimizer_runs = 1000000
+optimizer_runs = 100
```
snreynolds, line 6: crying

snreynolds (approved):

PR #60, opened 2025-04-01
Separate key management

zhongeric:


src/KeyManagement.sol
```diff
@@ -0,0 +1,67 @@
...
+/// @dev A base contract for managing keys.
+abstract contract KeyManagement is IKeyManagement {
+    using EnumerableSetLib for EnumerableSetLib.Bytes32Set;
+    using KeyLib for Key;
+
+    /// @dev Must be overridden by the implementation
+    function _authorizeCaller() internal view virtual {}
+
+    /// @inheritdoc IKeyManagement
+    function authorize(Key memory key) external returns (bytes32 keyHash) {
```
snreynolds, line 18: should we rename? this is an overused term and I think of this action more as add?

Also with the new hooks, I wonder if we should have a function that also can add(key, validationModule)

zhongeric: #61

PR #64, opened 2025-04-01
Refactor hooks

zhongeric:
- Add HookLib
- Remove VerifySignature hook
- Add basic test


src/interfaces/IHook.sol
```diff
@@ -2,14 +2,10 @@
 pragma solidity ^0.8.0;
 
 import {PackedUserOperation} from "account-abstraction/interfaces/PackedUserOperation.sol";
+import {IERC1271} from "./IERC1271.sol";
```
snreynolds, line 5: hm so a hook doesnt need to have any of these interfaces tho

snreynolds: i guess if we want this to define every possible entrypoint it should also have the before/after execute? is that in a sep pr?

snreynolds (commented):
What are your thoughts on setting a key WITH a hook? Are you thinking it would just be bundled together in a multicall?

src/MinimalDelegation.sol
```diff
@@ -66,17 +66,9 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
...
         (bytes32 keyHash, bytes calldata signature) = wrappedSignature.unwrap();
-        IHook hook = HookLib.get(keyHash, HookFlags.VERIFY_SIGNATURE);
-
-        bool isValid;
-        if (address(hook) != address(0)) {
-            isValid = hook.verifySignature(_hashTypedData(executeStruct.hash()), wrappedSignature);
-        } else {
-            // Use default signature verification.
-            isValid = _verifySignature(_hashTypedData(executeStruct.hash()), keyHash, signature);
+        if (!_verifySignature(_hashTypedData(executeStruct.hash()), keyHash, signature)) {
```
snreynolds, line 69: nit: but if we use it like this I think we should have a similar isValid name for better readability

then this is more clear that its "if NOT valid"

snreynolds (approved):

PR #67, opened 2025-04-02
Use executeUserOp

snreynolds:


snreynolds:
this is how i feel working with 4337 

cookies-inside-out

src/MinimalDelegation.sol
```diff
@@ -52,15 +52,24 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
...
+    /// @dev This function is executeable only by the EntryPoint contract, and is the main pathway for UserOperations to be executed.
+    /// UserOperations can be executed through the execute function, but another method of authorization (ie through a passed in signature) is required.
+    /// userOp.callData is abi.encodeCall(executeUserOp.selector, (bytes32 mode, bytes executionData)) where executionData is abi.encode(Call[]).
+    function executeUserOp(PackedUserOperation calldata userOp, bytes32) external onlyEntryPoint {
+        // Parse the keyHash from the signature. This is the keyHash that has been pre-validated as the correct signer over the UserOp data
+        // and must be used to check further on-chain permissions over the call execution.
+        // TODO: Handle keyHash authorization.
+        // (bytes32 keyHash,) = userOp.signature.unwrap();
+
+        // The only valid modes for executeUserOp are BATCHED_CALL and BATCHED_CALL_CAN_REVERT.
```
dianakocsis, line 64: does this need to be checked?

src/libraries/CalldataDecoder.sol
```diff
@@ -17,6 +17,14 @@ library CalldataDecoder {
     /// error SliceOutOfBounds();
     uint256 constant SLICE_ERROR_SELECTOR = 0x3b99b53d;
 
+    /// @notice Removes the selector from the calldata and returns the encoded params.
+    function removeSelector(bytes calldata data) internal pure returns (bytes calldata params) {
```
gretzke, line 21: should you check that data.length > 4 here?
or can this be trusted?

gretzke (approved):

src/MinimalDelegation.sol
snreynolds: Hm, we could but not sure its worth. I more so meant to comment this to point out that only the EXEC_TYPE is extracted from the mode.

src/libraries/CalldataDecoder.sol
snreynolds: Yeah thats a good call, it shouldn't be because it would fail before this assuming the entrypoint code doesn't change, but probably a good addition?

snreynolds: ah yeah so data.length is a full word btw so idt this is possible

zhongeric (approved):

PR #69, opened 2025-04-03
remove isSuperAdmin

dianakocsis:


snreynolds (approved):
approved! but have been having some thoughts around admin/permissions that I will push up which somewhat keeps admin behaviour, but I think defines it a bit differently and combines some logic with the hooks flow

PR #73, opened 2025-04-03
Add back hook.verifySignature

zhongeric:


test/utils/MockValidationHook.sol
```diff
@@ -5,9 +5,14 @@ import {IHook} from "src/interfaces/IHook.sol";
 import {PackedUserOperation} from "account-abstraction/interfaces/PackedUserOperation.sol";
 
 contract MockValidationHook is IHook {
+    bool internal _verifySignatureReturnValue;
     bytes4 internal _isValidSignatureReturnValue;
     uint256 internal _validateUserOpReturnValue;
```
zhongeric, line 10: I dislike how each of these returns a different value, I wonder if we should have some middleware or something, because you technically shouldn't be implementing 1271 on the hook

src/MinimalDelegation.sol
```diff
@@ -79,12 +79,19 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
         (uint256 nonce, bytes calldata wrappedSignature) = opData.decodeUint256Bytes();
```
snreynolds, line 79: nit/future: we should rename this?

snreynolds (approved):

PR #75, opened 2025-04-03
Pin 4337 entrypoint address as constant

### LINKED ISSUE #70: Pin 4337 entrypoint
Because we now have `executeUserOp` function, tied to current entrypoint implementation, safest to pin it

zhongeric:
Fix #70


src/MinimalDelegation.sol
```diff
@@ -168,11 +155,6 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
         return _1271_INVALID_VALUE;
     }
 
-    /// @inheritdoc IERC4337Account
-    function ENTRY_POINT() public view override returns (address) {
```
snreynolds, line 172: I think we should keep this function/keep update functionality

src/ERC4337Account.sol
```diff
@@ -24,4 +24,15 @@ abstract contract ERC4337Account is IERC4337Account {
     }
 
     function ENTRY_POINT() public view virtual returns (address);
+
+    /// @notice Packs the entry point into a uint256.
+    /// @dev Set the most significant bit to 1 to indicate that the entrypoint has been set by the user.
+    function _packEntryPoint(address entryPoint) internal pure returns (uint256) {
```
snreynolds, line 30: put in library?

entrypoint.pack()

entrypoint.isSet()

snreynolds: or maybe entrypoint.isOverridden() is more intuitive here? tells what the "set" flag does

zhongeric: just added!

src/MinimalDelegation.sol
```diff
@@ -112,10 +115,16 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
     /// @inheritdoc IERC4337Account
     function updateEntryPoint(address entryPoint) external {
         _onlyThis();
-        MinimalDelegationStorageLib.get().entryPoint = entryPoint;
+        MinimalDelegationStorageLib.get().entryPoint = EntrypointLib.pack(entryPoint);
```
snreynolds, line 118: nit: but I prefer libraries are used like
`entryPoint.pack()`

snreynolds (approved):

src/MinimalDelegation.sol
zhongeric: hmm yeah but we have different types here, the packed is 256 while we're wrapping an address here

snreynolds: what do you mean? `entryPoint` is an address

so you can do `using EntrypointLib for address` no?

PR #76, opened 2025-04-03
move functions into base contract

dianakocsis:


src/NonceManager.sol
```diff
@@ -7,11 +7,20 @@ import {MinimalDelegationStorage, MinimalDelegationStorageLib} from "./libraries
...
-    function getNonce(uint256 key) public view virtual returns (uint256 nonce);
+    function getNonce(uint256 key) public view override returns (uint256 nonce) {
+        return MinimalDelegationStorageLib.get().nonceSequenceNumber[uint192(key)] | (key << 64);
+    }
 
     /// @inheritdoc INonceManager
-    function invalidateNonce(uint256 nonce) public virtual;
+    function invalidateNonce(uint256 nonce) public override {
+        _onlyThis();
+        _invalidateNonce(nonce);
```
zhongeric, line 21: if _invalidateNonce isn't expected to be used anywhere else can just incorporate into one function

zhongeric (approved):

PR #78, opened 2025-04-03
Refactor hook callsites, Use key settings

snreynolds:


src/KeyManagement.sol
```diff
@@ -16,10 +17,24 @@ abstract contract KeyManagement is IKeyManagement {
...
 
     /// @inheritdoc IKeyManagement
-    function authorize(Key memory key) external returns (bytes32 keyHash) {
+    function register(Key memory key) external {
         _onlyThis();
-        keyHash = _authorize(key);
-        emit Authorized(keyHash, key);
+
+        bytes32 keyHash = key.hash();
+        // If the keyHash already exists, it does not revert and updates the key instead.
```
zhongeric, line 24: Can just early return if `keyHashes.contains(keyHash)`?

test/MinimalDelegation.isValidSignature.t.sol
```diff
@@ -127,8 +130,9 @@ contract MinimalDelegationIsValidSignatureTest is DelegationHandler, HookHandler
         TestKey memory p256Key = TestKeyManager.initDefault(KeyType.P256);
         bytes32 keyHash = p256Key.toKeyHash();
 
-        vm.prank(address(signerAccount));
-        signerAccount.setHook(keyHash, mockValidationHook);
+        vm.startPrank(address(signerAccount));
+        signerAccount.register(p256Key.toKey());
+        signerAccount.update(keyHash, Settings.wrap(0).fromHook(mockValidationHook));
```
zhongeric, line 135: nit: would prefer something like `Settings.init().withHook(mockValidationHook)`

src/KeyManagement.sol
snreynolds: this is more expensive in the register case where it hasn't been registered yet though

zhongeric (approved):

PR #80, opened 2025-04-03
Use custom storage lang support in solidity post 0.8.29

zhongeric:


src/KeyManagement.sol
```diff
@@ -3,15 +3,22 @@ pragma solidity ^0.8.23;
...
 import {IHook} from "./interfaces/IHook.sol";
 
-/// @dev A base contract for managing keys.
-abstract contract KeyManagement is IKeyManagement {
+struct KeyExtraStorage {
+    IHook hook;
+}
+
+/// @dev A base contract for managing keys
+contract KeyManagement is IKeyManagement {
```
zhongeric, line 14: I'm unsure of whether custom storage applies to abstract contracts further down in the inheritance tree based on this comment in the spec:

> The storage layout cannot be specified for abstract contracts, interfaces and libraries. Also, it is important to note that it does not affect transient state variables.

But maybe it means you can't declare it on an abstract contract? Since it has to be declared at the topmost contract in the inheritance

snreynolds: Hm that is interesting - worth leaving a comment about it in the forum I think, but I imagine that means it can't be declared on abstract contracts?

src/MinimalDelegation.sol
```diff
@@ -27,7 +27,10 @@ import {IHook} from "./interfaces/IHook.sol";
 import {SignatureUnwrapper} from "./libraries/SignatureUnwrapper.sol";
 import {HooksLib} from "./libraries/HooksLib.sol";
 
-contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receiver, KeyManagement, NonceManager {
+/// @notice Uses custom storage layout according to ERC7201
+/// @custom:storage-location erc7201:Uniswap.MinimalDelegation.1.0.0
+/// @dev keccak256(abi.encode(uint256(keccak256("Uniswap.MinimalDelegation.1.0.0")) - 1)) & ~bytes32(uint256(0xff))
+contract MinimalDelegation layout at 0xc807f46cbe2302f9a007e47db23c8af6a94680c1d26280fb9582873dbe5c9200 is IERC7821, ERC1271, EIP712, ERC4337Account, Receiver, KeyManagement, NonceManager, ERC7201 {
```
snreynolds, line 33: 🤣

src/KeyManagement.sol
zhongeric: yeah i think so too

test/MinimalDelegation.storage.t.sol
```diff
@@ -0,0 +1,72 @@
...
+        assertEq(readSlots.length, 1);
+        assertEq(writeSlots.length, 0);
+
+        bytes32 mappingRootSlot = _addOffset(signerAccount.CUSTOM_STORAGE_ROOT(), NONCE_SEQUENCE_NUMBER_SLOT);
+        bytes32 nestedSlot = _calculateNestedMappingSlot(key, mappingRootSlot);
+        assertEq(readSlots[0], nestedSlot);
+    }
+
+    function test_entrypoint() public {
+        vm.record();
```
snreynolds, line 65: nice

src/MinimalDelegation.sol
```diff
@@ -38,7 +41,9 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
     using ExecutionDataLib for ExecutionData;
     using SignatureUnwrapper for bytes;
     using HooksLib for IHook;
-    using EntrypointLib for uint256;
+    using EntrypointLib for *;
```
snreynolds, line 44: niceee

src/NonceManager.sol
```diff
@@ -2,32 +2,33 @@
...
 abstract contract NonceManager is INonceManager {
+    mapping(uint256 key => uint256 seq) nonceSequenceNumber;
+
     /// @dev Must be overridden by the implementation
     function _onlyThis() internal view virtual {}
 
     /// @inheritdoc INonceManager
     function getNonce(uint256 key) external view override returns (uint256 nonce) {
-        return MinimalDelegationStorageLib.get().nonceSequenceNumber[uint192(key)] | (key << 64);
+        return nonceSequenceNumber[key] | (key << 64);
```
snreynolds, line 16: hm this is not super easy to understand and I'm realizing now that the documentation is incorrect here.

the `key` parameter is a uint256 but only the lower 192 bits are used since it is just a key, so we can directly use the uint256 value, but I think we should mask the upper 64 bits before keying the mapping.

then we have to shift the key over 64 to return a nonce in the format of [key][nonceNumber] packed.

Maybe we can use some consistency in naming in the nonce manager to understand what values we are working with

key - 192 bits
packedNonce - 256 bits consists of key and nonceNumber
nonceNumber - 64 bits

zhongeric: ah yeah the logic must have gotten mixed up in handling merge conflicts. I can update the naming

zhongeric: wait i thought only the upper 192 bits are used for the key and the rest is the seq?

snreynolds: the upper 192 bits are used for the key when it is passed in as a packed nonce [key]|[seq]

here it is just being passed as a key, so its a uint256 but only the lower 192 bits are used, so we need to mask the upper bits

snreynolds: also i think we should change this to just return the sequence

snreynolds (approved):

PR #90, opened 2025-04-04
Tests overhaul

zhongeric:
- Add signing and packing helper functions for execute paths
- Add invariant tests
- Add builders for encoding self calls, generating random fuzz inputs


test/MinimalDelegation.execute.invariant.t.sol
```diff
@@ -0,0 +1,245 @@
...
+
+    IMinimalDelegation public signerAccount;
+
+    constructor(SetupParams memory _params) FunctionCallGenerator(_params._tokenA, _params._tokenB) {
+        signerAccount = _params._signerAccount;
+        for (uint256 i = 0; i < _params._keys.length; i++) {
+            keys.push(_params._keys[i]);
+        }
+        callers = _params._callers;
+        tokenA = ERC20Mock(_params._tokenA);
```
snreynolds, line 63: could use TokenHandler?

test/utils/FunctionCallGenerator.sol
```diff
@@ -0,0 +1,177 @@
...
+    function _getSettingsForKeyHash(bytes32 keyHash) internal virtual returns (Settings) {}
+
+    /**
+     * @notice Generate a random function call with equal weighting between function types
+     * @param randomSeed Random seed for generation
+     * @param depth Current recursion depth
+     * @return A call object for the generated function
+     */
+    function generateRandomFunctionHandlerCall(uint256 randomSeed, uint256 depth) public returns (HandlerCall memory) {
+        vm.assume(randomSeed < type(uint256).max);
```
snreynolds, line 102: Why do you need to add this line?

test/utils/FunctionCallGenerator.sol
```diff
@@ -0,0 +1,177 @@
...
+        // Select function type with equal weighting
+        uint256 functionType = _bound(randomSeed, 0, FUNCTION_COUNT - 1);
+
+        // REGISTER
+        if (functionType == FUNCTION_REGISTER || _ghostKeyHashes.values().length == 0) {
+            TestKey memory newKey = _randomTestKey();
+            if (newKey.toKeyHash() != bytes32(0)) {
+                return _registerCall(newKey);
```
snreynolds, line 110: technically you could register the 0 address key, I wonder if we should test that edge case and make sure that registering the 0 address key with restricted settings, doesnt actually do anything

test/utils/CallUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+    using TestKeyManager for TestKey;
+
+    // Constants
+    address constant SELF_CALL = address(0);
+    bytes32 internal constant BATCHED_CALL = 0x0100000000000000000000000000000000000000000000000000000000000000;
+
+    // Call array operations
+
+    /// @dev Create empty Call array
+    function init() internal pure returns (Call[] memory) {
```
snreynolds, line 31: nit: maybe we have initCall and initCallArray?

test/utils/CallUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    // Call encoding for specific operations
+
+    /// @dev Add execute call to array
+    function addExecute(Call[] memory calls, Call[] memory innerCalls) internal pure returns (Call[] memory) {
+        return push(calls, encodeExecuteCall(innerCalls));
+    }
+
+    /// @dev Create call to execute a batch
+    function encodeExecuteCall(Call[] memory calls) internal pure returns (Call memory) {
```
snreynolds, line 75: it would be nice if this could handle different modes i think

snapshots/MinimalDelegationExecuteTest.json
```diff
@@ -1,10 +1,10 @@
 {
-  "execute_BATCHED_CALL_SUPPORTS_OPDATA_singleCall": "92418",
-  "execute_BATCHED_CALL_SUPPORTS_OPDATA_twoCalls": "100114",
+  "execute_BATCHED_CALL_SUPPORTS_OPDATA_singleCall": "93054",
```
snreynolds, line 2: hmm these shouldnt be changing I think

test/utils/ExecuteHandler.sol
```diff
@@ -1,13 +1,46 @@
...
+        bytes memory signature = _sign(digest, key);
+        packedSignature = abi.encode(nonce, abi.encode(keyHash, signature));
+    }
+
+    /// @notice Base function for signing calls
+    /// @dev Assume that the digest is either:
+    /// - SignedCalls if using internal verification
+    /// - Wrapped if using 1271
+    /// - The userOpHash if using ERC-4337
+    function _sign(bytes32 digest, TestKey memory key) internal view returns (bytes memory signature) {
```
snreynolds, line 43: nit: i think just doing key.sign(digest) inline instead of calling another internal _sign function is totally fine

less paths to do the same thing

test/utils/ExecuteHandler.sol
```diff
@@ -1,13 +1,46 @@
...
+    bytes32 internal constant ROOT_KEY_HASH = bytes32(0);
+
+    /// @notice Sign calls and pack them for internal verification path
+    function _signAndPack(bytes32 digest, TestKey memory key, uint256 nonce) internal view returns (bytes memory) {
+        bytes32 keyHash = key.toKeyHash();
+        return _signAndPack(digest, key, nonce, keyHash);
+    }
+
+    /// @dev If the signing key is the root EOA, pass bytes32(0) for keyHash
+    function _signAndPack(bytes32 digest, TestKey memory key, uint256 nonce, bytes32 keyHash)
```
snreynolds, line 29: why do we need both of these functions? seems like _signAndPack can just take a Key type

test/utils/TestKeyManager.sol
```diff
@@ -41,11 +41,7 @@ library TestKeyManager {
             return TestKey({keyType: keyType, publicKey: abi.encodePacked(x, y), privateKey: DEFAULT_SECP256R1_PK});
         } else if (keyType == KeyType.Secp256k1) {
             address defaultAddress = vm.addr(DEFAULT_SECP256K1_PK);
-            return TestKey({
-                keyType: keyType,
-                publicKey: abi.encodePacked(defaultAddress),
-                privateKey: DEFAULT_SECP256K1_PK
-            });
+            return TestKey({keyType: keyType, publicKey: abi.encode(defaultAddress), privateKey: DEFAULT_SECP256K1_PK});
```
snreynolds, line 44: I think some files formatting is messed up, I'm gonna make a PR to move out the layout syntax so we can start to run forge fmt locally

snreynolds: & nice - this should be abi.encode not abi.encodePacked for sure

test/utils/CallUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    function withData(Call memory call, bytes memory data) internal pure returns (Call memory) {
+        call.data = data;
+        return call;
+    }
+
+    // Call encoding for specific operations
+
+    /// @dev Add execute call to array
+    function addExecute(Call[] memory calls, Call[] memory innerCalls) internal pure returns (Call[] memory) {
```
snreynolds, line 70: high level comment, but its unclear to me which should be used from this library

addExecute or encodeExecuteCall ? same with the other functions below bc theres essentially 2 lib functions per action.

I feel like a better abstraction would be to expose building the specific call and then just having a generic push function that way we can be a bit more granular in the tests for what "kind" of execute or revoke call we want to create

test/NonceManager.t.sol
```diff
@@ -73,7 +73,7 @@ contract NonceManagerTest is DelegationHandler {
 
     function test_fuzz_invalidateNonce(uint256 nonceKey, uint16 sequence) public {
         // Skip sequences that would overflow when incremented
-        sequence = uint16(bound(sequence, 1, type(uint16).max));
+        sequence = uint16(_bound(sequence, 1, type(uint16).max));
```
snreynolds, line 76: why the switch to internal bound?

test/MinimalDelegation.execute.t.sol
```diff
@@ -445,76 +406,86 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
     /// forge-config: ci.isolate = true
     function test_execute_native_single_batchedCall_opData_eoaSigner_gas() public {
-        Call[] memory calls = CallBuilder.init();
+        Call[] memory calls = CallUtils.init();
         calls = calls.push(buildTransferCall(address(0), address(receiver), 1e18));
 
-        uint256 nonceKey = 0;
-        (uint256 nonce,) = _buildNextValidNonce(nonceKey);
-        SignedCalls memory signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        bytes memory executionData = abi.encode(calls, _signAndPack(hash(calls, DEFAULT_NONCE), signerTestKey, DEFAULT_NONCE, ROOT_KEY_HASH));
```
snreynolds, line 412: i think this makes it harder to understand how data is encoded for the contracts imo. like i honestly forget where the nonce goes in the opdata structuring bc there are multiple things being packed right? the signature and then the nonce?

Maybe it would help to just separate out a few steps?

wrappedSignature = _signAndPack(key, signedCalls) // note the nonce is already passed in
opData = pack(nonce, wrappedSignature)
executionData = encode(calls, opData)

test/MinimalDelegation.execute.t.sol
```diff
@@ -176,48 +177,57 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
-        bytes32 hashToSign = signerAccount.hashTypedData(signedCalls.hash());
-        (uint8 v, bytes32 r, bytes32 s) = vm.sign(signerPrivateKey, hashToSign);
-        bytes memory signature = abi.encodePacked(r, s, v);
-
-        // Pack the execution data:
-        // 1. Encode the nonce and signature into opData
-        bytes memory opData = abi.encode(nonce, signature);
-        // 2. Encode the calls and opData together
-        bytes memory executionData = abi.encode(calls, opData);
+        bytes memory executionData = abi.encode(calls, _signAndPack(hash(calls, DEFAULT_NONCE), signerTestKey, nonce, ROOT_KEY_HASH));
```
snreynolds, line 229: strange that this changed the gas costs for snapshotGasLastCall...

test/MinimalDelegation.execute.t.sol
```diff
@@ -445,76 +406,86 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        TestKey memory newKey = TestKeyManager.initDefault(KeyType.Secp256k1);
+        Call[] memory calls = CallUtils.init();
+        calls = CallUtils.addRegister(calls, newKey);
+        calls = CallUtils.addUpdate(calls, newKey.toKeyHash(), Settings.wrap(0));
+
+        vm.expectRevert(IERC7821.Unauthorized.selector);
+        signerAccount.execute(BATCHED_CALL, abi.encode(calls));
+    }
+
+    function test_execute_register_update_withRootSignature_succeeds() public {
```
snreynolds, line 481: initDefault isn't the root key though ?

test/MinimalDelegation.execute.t.sol
```diff
@@ -445,76 +406,86 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        TestKey memory newKey = TestKeyManager.initDefault(KeyType.Secp256k1);
+        Call[] memory calls = CallUtils.init();
+        calls = CallUtils.addRegister(calls, newKey);
+        calls = CallUtils.addUpdate(calls, newKey.toKeyHash(), Settings.wrap(0));
+
+        bytes memory executionData = abi.encode(calls, _signAndPack(hash(calls, DEFAULT_NONCE), signerTestKey, DEFAULT_NONCE, ROOT_KEY_HASH));
 
-        // Execute the batch of calls with the signature
-        vm.startPrank(address(signerAccount));
         signerAccount.execute(BATCHED_CALL_SUPPORTS_OPDATA, executionData);
```
snreynolds, line 489: this should assert keyHashes.contains(keyHash) with settings of 0?

test/MinimalDelegation.execute.invariant.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        targetSender(sender);
+    }
+
+    /// @notice Verifies that the root key can always revoke other keys
+    function invariant_rootKeyCanAlwaysRevokeOtherSigningKeys() public {
+        // Ensure key exists
+        bytes32 keyHash = untrustedKey.toKeyHash();
+        try signerAccount.getKey(keyHash) {
+            vm.prank(address(signerAccount));
+            signerAccount.revoke(keyHash);
```
snreynolds, line 221: shouldn't this assert that the revoke has actually removed the state from the contract and/or another call from that key is invalid?

snreynolds: also how would we know if the state ever catches the first part of the try? ie that the specific untrusted key hash ever existed on the contract?

test/NonceManager.t.sol
zhongeric: doesn't log

test/utils/FunctionCallGenerator.sol
zhongeric: bc the recursion below does seed + 1 for the recursive call

test/utils/CallUtils.sol
zhongeric: Yea its tricky because other modes require opData, which is encoded differently

test/utils/ExecuteHandler.sol
zhongeric: KeyType doesn't tell us if its the root k1 key or a different one

test/MinimalDelegation.execute.t.sol
zhongeric: That makes sense, my only point is that it takes so much additional code to write a passing test right now because of how specific the encoding is (maybe that's a meta point for the contracts lol), but these helper functions were meant to be used in tests where we're not explicitly testing something related to the encoding

test/MinimalDelegation.execute.t.sol
zhongeric: good catch

zhongeric: oh its because the root key is calling the functions to register the newKey and update it

test/utils/CallUtils.sol
zhongeric: Yeah makes sense, I was struggling with the abstractions here, was too deep in the test rabbit hole

zhongeric: Going to remove the unused funcs for now

test/MinimalDelegation.execute.invariant.t.sol
zhongeric: agree that it should assert it removed the key from the contract state but the `assert` in the catch branch checks that the only reason it should revert is if the key does not exist (it was probably removed by the fuzzed invariant test)

test/utils/ExecuteHandler.sol
zhongeric: removed though

test/MinimalDelegation.execute.t.sol
zhongeric: OK I simplified it!

test/MinimalDelegation.execute.t.sol
zhongeric: should be fixed now, I was adding bytes32(0) to the root key when signing which wasn't what we were doing on main

test/MinimalDelegation.execute.invariant.t.sol
zhongeric: I wanted to pass the tokens in by reference here, so we can change the state within the invariant handler and test for it outside in the test file

test/MinimalDelegation.execute.invariant.t.sol
zhongeric: removed these because `untrustedSigningKey` will never be registered on the account now

test/MinimalDelegation.execute.invariant.t.sol
```diff
@@ -0,0 +1,244 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.23;
+
+import {console2} from "forge-std/console2.sol";
```
snreynolds, line 4: rm?

snreynolds: nm! I see its how its being used

test/utils/FunctionCallGenerator.sol
snreynolds: seems like this can be removed now?

zhongeric: yes!

test/utils/FunctionCallGenerator.sol
zhongeric: yes I think this is better as an explicit test case

test/MinimalDelegation.execute.invariant.t.sol
```diff
@@ -0,0 +1,245 @@
...
+    using WrappedDataHash for bytes32;
+
+    MinimalDelegationExecuteInvariantHandler internal invariantHandler;
+
+    address public untrustedCaller = makeAddr("untrustedCaller");
+    uint256 public untrustedPrivateKey = 0xdead;
+
+    address internal sender = makeAddr("sender");
+
+    TestKey[] internal _keys;
```
snreynolds, line 172: interesting so the _keys array here is really just used to pull a signer, but it seems like it would never test the state where we have created a new key (in _generate_generateHandlerCall) and then use that newly registered key to sign execution

test/utils/FunctionCallGenerator.sol
```diff
@@ -0,0 +1,160 @@
...
+        return abi.encodeWithSelector(IERC7821.CallFailed.selector, abi.encodePacked(selector));
+    }
+
+    /// @return calldata to register a new key along with its callback
+    function _registerCall(TestKey memory newKey, bytes memory revertData)
+        internal
+        virtual
+        returns (HandlerCall memory)
+    {
+        // No error is thrown if the key is already registered, so ignore
```
snreynolds, line 70: there is an error thrown though if you try to register a key that is the signer account

snreynolds: i believe that won't happen here since the newKey is randomly generated though right?

snreynolds: (also same for update call)

snreynolds (commented):
would just be curious to get an idea of the test coverage that running these tests hit? some logs would be fine and we can comment them out

test/utils/ExecuteHandler.sol
```diff
@@ -1,13 +1,26 @@
...
         0x0101000000000000000000000000000000000000000000000000000000000000;
     bytes32 internal constant BATCHED_CALL_SUPPORTS_OPDATA =
         0x0100000000007821000100000000000000000000000000000000000000000000;
     bytes32 internal constant BATCHED_CALL_SUPPORTS_OPDATA_AND_CAN_REVERT =
         0x0101000000007821000100000000000000000000000000000000000000000000;
+
+    uint256 internal constant DEFAULT_NONCE = 0;
+
+    /// @notice Generate a signature over the digest and pack it with the keyHash
+    function _signAndPack(bytes32 digest, TestKey memory key) internal pure returns (bytes memory packedSignature) {
```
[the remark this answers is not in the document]
zhongeric: ah yes I removed

test/MinimalDelegation.execute.invariant.t.sol
zhongeric: Yea, I can explore that in a follow up I think

test/utils/FunctionCallGenerator.sol
zhongeric: Ah yeah I guess you would need a collision

test/MinimalDelegation.execute.invariant.t.sol
snreynolds: yeah I more so called it out cause I was confused by the different key states.. there is like fixture keys which I believe is just random keys that might get added/removed to the account and then keys which is just initialized with 2 to begin with (one root, one unauthorized) and those would actually sign new information

it might be helpful to separate them out by like signingKeys toRegisterKeys or something? and keys that are registered should be added to the signingKeys imo

test/utils/FunctionCallGenerator.sol
snreynolds: i think maybe add a case in the case that we end up combining key state and the root key is an option as an input into these function generators

test/MinimalDelegation.execute.t.sol
```diff
@@ -750,4 +744,58 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+
+        vm.prank(address(signerAccount));
+        signerAccount.execute(BATCHED_CALL, abi.encode(calls));
+    }
+
+    function test_execute_register_update_asNonRoot_reverts() public {
+        TestKey memory newKey = TestKeyManager.initDefault(KeyType.Secp256k1);
+        Call[] memory calls = CallUtils.initArray();
+        calls = calls.push(CallUtils.encodeRegisterCall(newKey));
+        calls = calls.push(CallUtils.encodeUpdateCall(newKey.toKeyHash(), Settings.wrap(0)));
```
snreynolds, line 780: this would be interesting if we sent these in with the mode BATCHED_CALL_CAN_REVERT and expect no state to change but no revert to happen

test/utils/ExecuteHandler.sol
```diff
@@ -1,8 +1,13 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.23;
 
+import {Call} from "../../src/libraries/CallLib.sol";
```
snreynolds, line 4: unused

snreynolds (approved):

test/MinimalDelegation.execute.t.sol
zhongeric: Ooo good idea

zhongeric: Ah so the Unauthorized is thrown during validation not execution so the mode isn't applied

test/utils/FunctionCallGenerator.sol
zhongeric: added in new PR

PR #92, opened 2025-04-04
feature(custom-storage): fix nonce manager to use getSeq

zhongeric:


src/NonceManager.sol
```diff
@@ -12,8 +12,8 @@ abstract contract NonceManager is INonceManager {
     function _onlyThis() internal view virtual {}
 
     /// @inheritdoc INonceManager
-    function getNonce(uint256 key) external view override returns (uint256 nonce) {
-        return nonceSequenceNumber[key] | (key << 64);
+    function getSeq(uint256 key) external view override returns (uint256 seq) {
+        return nonceSequenceNumber[key];
```
snreynolds, line 16: mask upper 64?

PR #95, opened 2025-04-04
Batch of batches

dianakocsis:


zhongeric (commented):
looks good just comment abt the mode

src/MinimalDelegation.sol
```diff
@@ -64,6 +65,11 @@ contract MinimalDelegation is
             (Call[] calldata calls, bytes calldata opData) = executionData.decodeCallsBytes();
             _authorizeOpData(mode, calls, opData);
             _dispatch(mode, calls);
+        } else if (mode.isBatchOfBatches()) {
+            bytes[] calldata executeDataArray = executionData.decodeBytesArray();
```
snreynolds, line 69: use memory, we are scrapping calldata decoder for now

test/MinimalDelegation.execute.t.sol
```diff
@@ -517,4 +517,105 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        uint256 nonceKey = 0;
+        (uint256 nonce,) = _buildNextValidNonce(nonceKey);
+
+        // Create hash of the calls + nonce and sign it
+        SignedCalls memory signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        bytes32 hashToSign = signerAccount.hashTypedData(signedCalls.hash());
+        bytes memory signature = signerTestKey.sign(hashToSign);
+        // Pack the execution data:
+        // 1. Encode the nonce and signature into opData
+        bytes memory opData = abi.encode(nonce, signature);
```
snreynolds, line 535: this signature needs to be wrapped with the keyHash

test/MinimalDelegation.execute.t.sol
```diff
@@ -517,4 +517,105 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        calls = calls.push(buildTransferCall(address(tokenA), address(receiver), 1e18)); // Transfer 1 tokenA
+
+        // Get the current nonce components for key 0
+        uint256 nonceKey = 0;
+        (uint256 nonce,) = _buildNextValidNonce(nonceKey);
+
+        // Create hash of the calls + nonce and sign it
+        SignedCalls memory signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        bytes32 hashToSign = signerAccount.hashTypedData(signedCalls.hash());
+        bytes memory signature = signerTestKey.sign(hashToSign);
```
snreynolds, line 532: can we actually test it with a signature that is not the root signer?

test/MinimalDelegation.execute.t.sol
```diff
@@ -517,4 +517,105 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+
+        // Create hash of the calls + nonce and sign it
+        nonce++; // increment the nonce since first one hasnt been used yet
+        signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        hashToSign = signerAccount.hashTypedData(signedCalls.hash());
+        signature = signerTestKey.sign(hashToSign);
+
+        // Pack the execution data:
+        // 1. Encode the nonce and signature into opData
+        opData = abi.encode(nonce, signature);
```
snreynolds, line 550: same need to wrap the signature

snreynolds: also for the second batch would be sweet if this tested on a different key than the first call

test/MinimalDelegation.execute.t.sol
```diff
@@ -517,4 +517,105 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        (uint256 nonce,) = _buildNextValidNonce(nonceKey);
+
+        // Create hash of the calls + nonce and sign it
+        SignedCalls memory signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        bytes32 hashToSign = signerAccount.hashTypedData(signedCalls.hash());
+        bytes memory signature = signerTestKey.sign(hashToSign);
+
+        // Pack the execution data:
+        // 1. Encode the nonce and signature into opData
+        bytes memory opData = abi.encode(nonce, signature);
```
snreynolds, line 584: wrap sig

test/MinimalDelegation.execute.t.sol
```diff
@@ -517,4 +517,105 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+
+        // Create hash of the calls + nonce and sign it
+        nonce++; // increment the nonce since first one hasnt been used yet
+        signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        hashToSign = signerAccount.hashTypedData(signedCalls.hash());
+        signature = signerTestKey.sign(hashToSign);
+
+        // Pack the execution data:
+        // 1. Encode the nonce and signature into opData
+        opData = abi.encode(nonce, signature);
```
snreynolds, line 600: & anotha wrap :)

this is tricky bc our calldata decoder doesnt revert rn... but this willl revert when u pull in main

snapshots/MinimalDelegationExecuteTest.json
```diff
@@ -1,12 +1,3 @@
...
-  "execute_BATCHED_CALL_SUPPORTS_OPDATA_twoCalls": "131689",
-  "execute_BATCHED_CALL_opData_P256_singleCall": "117579",
-  "execute_BATCHED_CALL_opData_singleCall": "94774",
-  "execute_BATCHED_CALL_opData_singleCall_native": "95672",
-  "execute_BATCHED_CALL_opData_twoCalls": "131689",
-  "execute_BATCHED_CALL_singleCall": "61185",
-  "execute_BATCHED_CALL_singleCall_native": "62081",
-  "execute_BATCHED_CALL_twoCalls": "97364",
-  "execute_invalidMode_reverts": "22940"
+  "execute_BATCH_OF_BATCHES_twoBatches_oneCallEach": "104674"
```
zhongeric, line 2: what happened to these gas snaps?

dianakocsis: woops

src/MinimalDelegation.sol
```diff
@@ -51,7 +52,7 @@ contract MinimalDelegation is
 
     uint256 public packedEntrypoint;
 
-    function execute(bytes32 mode, bytes calldata executionData) external payable override {
+    function execute(bytes32 mode, bytes memory executionData) public payable override {
```
snreynolds, line 55: oh interesting.. this essentially kills any kind of potential calldata decoding optimization. i think its fine but an alternative could be to dispatch internally

test/MinimalDelegation.execute.t.sol
```diff
@@ -592,4 +592,113 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        vm.prank(address(signerAccount));
+        signerAccount.register(webAuthnP256Key.toKey());
+
+        // Get the current nonce components for key 0
+        uint256 nonceKey = 0;
+        (uint256 nonce,) = _buildNextValidNonce(nonceKey);
+
+        // Create hash of the calls + nonce and sign it
+        SignedCalls memory signedCalls = SignedCalls({calls: calls, nonce: nonce});
+        bytes memory signature = abi.encode(
```
snreynolds, line 611: nit: name bytes memory executionData

test/MinimalDelegation.execute.t.sol
```diff
@@ -592,4 +592,113 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        bytes memory signature = signerTestKey.sign(hashToSign);
+
+        // Pack the execution data:
+        // 1. Encode the nonce and signature into opData
+        bytes memory opData = abi.encode(nonce, abi.encode(KeyLib.ROOT_KEY_HASH, signature));
+        // 2. Encode the calls and opData together
+        bytes memory executionData1 = abi.encode(calls, opData);
+
+        calls = CallBuilder.init();
+        // this call silently reverts because its over the balance
```
snreynolds, line 672: this doesn't silently revert though right? it does cause a revert

test/MinimalDelegation.execute.t.sol
```diff
@@ -592,4 +592,113 @@ contract MinimalDelegationExecuteTest is TokenHandler, DelegationHandler, Execut
...
+        bytes memory executionData2 = abi.encode(calls, signature);
+
+
+        bytes[] memory executionDataArray = new bytes[](2);
+        executionDataArray[0] = executionData1;
+        executionDataArray[1] = executionData2;
+
+        // execute the batch of batches
+        signerAccount.execute(BATCH_OF_BATCHES_CALL, abi.encode(executionDataArray));
+        vm.snapshotGasLastCall("execute_BATCH_OF_BATCHES_twoBatches_oneCallEach");
```
snreynolds, line 643: Can you move the gas snapshot to another test and then add the inline forge configs to it

test/MinimalDelegation.execute.t.sol
dianakocsis: yess forgot to update the comment

src/MinimalDelegation.sol
dianakocsis: the reason i changed it to memory was because when decoding batch of batches, it needs to call back into execute. so i think if we went back to calldata decoding, we would be able to switch it back to calldata then right?

snreynolds: correct if we decoded in calldata or if we dispatched it internally

snreynolds (approved):

src/MinimalDelegation.sol
dianakocsis: yeah so we prob would want to call back to `execute` so it follows the erc

PR #100, opened 2025-04-04
Add expiry logic and tests

zhongeric:


zhongeric:
fix #100

src/MinimalDelegation.sol
```diff
@@ -142,9 +143,11 @@ contract MinimalDelegation is IERC7821, ERC1271, EIP712, ERC4337Account, Receive
         _payEntryPoint(missingAccountFunds);
         (bytes32 keyHash, bytes calldata signature) = userOp.signature.unwrap();
 
-        IHook hook = keySettings[keyHash].hook();
+        Settings settings = keySettings[keyHash];
+        // TODO: should this be return 0 or revert
+        if (settings.isExpired()) revert IKeyManagement.KeyExpired();
```
snreynolds, line 148: I think revert is good

Its kinda funny that the settings can expire? feels like odd that its not really clear its the KEY that has expired?

src/libraries/SettingsLib.sol
```diff
@@ -34,4 +34,11 @@ library SettingsLib {
             _hook := and(settings, mask)
         }
     }
+
+    function isExpired(Settings settings) internal view returns (bool _isExpired) {
+        uint256 _exp = expiration(settings);
+        // Keys with expiry of 0 never expire
+        if (_exp == 0) return false;
+        return _exp < block.timestamp;
```
snreynolds, line 42: so its not inclusive? if you set a key to expire at 10, and it IS 10 does that mean it should be false?

zhongeric: just updated to be inclusive

snreynolds (approved):

PR #102, opened 2025-04-07
feat: add example spending limit hook

zhongeric:


src/hooks/execution/SpendingLimitHook.sol
```diff
@@ -0,0 +1,240 @@
...
+        /// @dev The start of the last updated period (unix timestamp).
+        uint256 lastUpdated;
+    }
+
+    /// @dev Holds the storage for the token spend limits.
+    struct TokenSpendStorage {
+        /// @dev An enumerable set of the periods.
+        EnumerableSetLib.Uint8Set periods;
+        /// @dev Mapping of `uint8(period)` to `TokenPeriodSpendStorage`.
+        mapping(uint256 => TokenPeriodSpendStorage) spends;
```
snreynolds, line 84: why is this BY period? How would that work if you have two spend periods for the same token that overlap? You'll have to check both periods states and combine the spends somehow?

src/hooks/execution/SpendingLimitHook.sol
```diff
@@ -0,0 +1,240 @@
...
+        uint256 totalNativeSpend;
+        address[] erc20s;
+        uint256[] transferAmounts;
+        uint256[] balancesBefore;
+    }
+
+    /// @dev Set the spend limit for a key hash.
+    /// Uses msg.sender to compute the accountKeyHash.
+    function setSpendLimit(bytes32 keyHash, address token, SpendPeriod period, uint256 limit) external {
+        AccountKeyHash accountKeyHash = keyHash.wrap(msg.sender);
```
snreynolds, line 107: maybe use a different word? wrap is keyword for a custom type

src/hooks/execution/SpendingLimitHook.sol
zhongeric: If you have two periods for the same token they all must be true:
- Minute, 10 AND hour, 100

You would loop through all of the periods, increment the spend for each and check none exceed?

zhongeric: You add the same amount to each tracked period spend

src/hooks/execution/SpendingLimitHook.sol
zhongeric: ok! I did `hashSender`

src/hooks/execution/SpendingLimitHook.sol
```diff
@@ -0,0 +1,245 @@
...
+            transferAmounts.p(LibBytes.loadCalldata(data, 0x24)); // `amount`.
+        }
+
+        // Sum transfer amounts, grouped by the ERC20s. In-place.
+        LibSort.groupSum(erc20s.data, transferAmounts.data);
+
+        // Collect the ERC20 balances before the batch execution.
+        for (uint256 i; i < erc20s.length(); ++i) {
+            address token = erc20s.getAddress(i);
+            balancesBefore.set(i, SafeTransferLib.balanceOf(token, msg.sender));
```
snreynolds, line 168: this could have duplicates no?

src/hooks/execution/SpendingLimitHook.sol
```diff
@@ -0,0 +1,245 @@
...
+        if (data.length < 4) {
+            return (IExecutionHook.beforeExecute.selector, abi.encode(tempStorage));
+        }
+
+        // We will only filter based on functions that are known to use `msg.sender`.
+        // For signature-based approvals (e.g. permit), we can't do anything
+        // to guard, as anyone else can directly submit the calldata and the signature.
+        uint32 fnSel = uint32(bytes4(LibBytes.loadCalldata(data, 0x00)));
+        // `transfer(address,uint256)`.
+        if (fnSel == 0xa9059cbb) {
```
snreynolds, line 157: why are we not doing a balance check?

src/hooks/execution/SpendingLimitHook.sol
```diff
@@ -0,0 +1,245 @@
...
+            if (tokenSpends.periods.length() == uint256(0)) continue;
+            _incrementSpent(
+                tokenSpends,
+                // While we can actually just use the difference before and after,
+                // we also want to let the sum of the transfer amounts in the calldata to be capped.
+                // This prevents tokens to be used as flash loans, and also handles cases
+                // where the actual token transfers might not match the calldata amounts.
+                // There is no strict definition on what constitutes spending,
+                // and we want to be as conservative as possible.
+                Math.max(
```
snreynolds, line 204: this is just odd to me... what if you like flash loaned something? it would still count it as a spend here even though your balance stayed under some limit this would count it incorrectly no?

snreynolds:
Closing as we may explore a simpler spending limit example :)

PR #103, opened 2025-04-07
use memory decoding, fix tests, harden key pathways

snreynolds:


src/KeyManagement.sol
```diff
@@ -33,6 +33,7 @@ abstract contract KeyManagement is IKeyManagement {
 
     function update(bytes32 keyHash, Settings settings) external {
         _onlyThis();
+        if (keyHash == KeyLib.ROOT_KEY_HASH) revert CannotUpdateRootKey();
```
zhongeric, line 36: I think we also have to check in register that you don't register the root key public key or something, because you could register it, and then it's `key.hash()` would return not bytes32(0), so you could techinically update it lol

test/utils/TestKeyManager.sol
```diff
@@ -101,6 +100,10 @@ library TestKeyManager {
     }
 
     function toKeyHash(TestKey memory key) internal pure returns (bytes32) {
+        /// If the key is the root user return a hash of 0 to signify the root user.
+        if (key.keyType == KeyType.Secp256k1 && abi.decode(key.publicKey, (address)) == signerTestKeyAddress) {
+            return bytes32(0);
+        }
```
zhongeric, line 106: im a little worried about footguns from this function being different than `KeyLib.hash`, could we just special case this inline in the tests?

zhongeric (commented):
nice! I wonder why some gas snaps went down but I'm generally very happy w this!

src/KeyManagement.sol
zhongeric: also this means we can't add hooks to root key

test/utils/TestKeyManager.sol
snreynolds: hmm what do you mean? this is a test function?

snreynolds: ok I changed this so we have to explicitly use bytes32(0) in the tests

snreynolds: I wonder if we should use not such a "default" value? Like it if should actually be some non zero value..

zhongeric (approved):

test/utils/TestKeyManager.sol
zhongeric: i don't mind doing in a follow up we should track

src/EIP712.sol
```diff
@@ -53,7 +53,7 @@ contract EIP712 is IEIP712, IERC5267 {
         (name, version) = _domainNameAndVersion();
         chainId = block.chainid;
         verifyingContract = address(this);
-        salt = salt; // `bytes32(0)`.
+        salt = salt; // `KeyLib.ROOT_KEY_HASH`.
```
zhongeric, line 56: nit: remove lol

test/ERC712.t.sol
```diff
@@ -41,7 +42,7 @@ contract ERC712Test is DelegationHandler, TokenHandler, FFISignTypedData {
         // Ensure that verifying contract is the signer
         assertEq(verifyingContract, address(signerAccount));
         assertEq(abi.encode(extensions), abi.encode(new uint256[](0)));
-        assertEq(salt, bytes32(0));
+        assertEq(salt, KeyLib.ROOT_KEY_HASH);
```
zhongeric, line 45: this too

test/utils/UserOpBuilder.sol
```diff
@@ -19,12 +20,12 @@ library UserOpBuilder {
 
     /// @dev Override initcode hash set to default of 0
     function encode(PackedUserOperation calldata userOp) internal pure returns (bytes memory) {
-        return userOp.encode(bytes32(0));
+        return userOp.encode(KeyLib.ROOT_KEY_HASH);
```
zhongeric, line 23: rm

test/ERC712.t.sol
```diff
@@ -14,6 +14,7 @@ import {TestKeyManager, TestKey} from "./utils/TestKeyManager.sol";
 import {TokenHandler} from "./utils/TokenHandler.sol";
 import {FFISignTypedData} from "./utils/FFISignTypedData.sol";
 import {SignedCallsLib, SignedCalls} from "../src/libraries/SignedCallsLib.sol";
+import {KeyLib} from "../src/libraries/KeyLib.sol";
```
zhongeric, line 17: rm

test/utils/UserOpBuilder.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.23;
 
 import {UserOperationLib} from "account-abstraction/core/UserOperationLib.sol";
 import {PackedUserOperation} from "account-abstraction/interfaces/PackedUserOperation.sol";
+import {KeyLib} from "../../src/libraries/KeyLib.sol";
```
zhongeric, line 6: rm

zhongeric (approved):

zhongeric (approved):

PR #107, opened 2025-04-07
add expiration in validationData returned from validateUserOp

dianakocsis:


src/MinimalDelegation.sol
```diff
@@ -151,26 +151,31 @@ contract MinimalDelegation is
 
         Settings settings = keySettings[keyHash];
         if (settings.isExpired()) revert IKeyManagement.KeyExpired();
```
snreynolds, line 153: Can isExpired return both the validity and the expiration? also we can then revert with the expiration timestamp which would be nice

and then you don't need the next line :)

src/MinimalDelegation.sol
```diff
@@ -151,26 +151,31 @@ contract MinimalDelegation is
...
+        uint40 expiry = settings.expiration();
 
         IHook hook = settings.hook();
         validationData = hook.hasPermission(HooksLib.VALIDATE_USER_OP_FLAG)
-            ? hook.validateUserOp(keyHash, userOp, userOpHash)
-            : _handleValidateUserOp(keyHash, signature, userOp, userOpHash);
+            ? hook.validateUserOp(keyHash, userOp, userOpHash, expiry)
```
snreynolds, line 158: I would actually say don't pass the expiry here, if the hook wants to check expiry they can call the getter function on the main delegated contract

src/MinimalDelegation.sol
```diff
@@ -151,26 +151,31 @@ contract MinimalDelegation is
 
         Settings settings = keySettings[keyHash];
         if (settings.isExpired()) revert IKeyManagement.KeyExpired();
+        uint40 expiry = settings.expiration();
 
         IHook hook = settings.hook();
         validationData = hook.hasPermission(HooksLib.VALIDATE_USER_OP_FLAG)
             ? hook.validateUserOp(keyHash, userOp, userOpHash)
```
zhongeric, line 158: should we pass expiry in here to the hook or is it something we should only use internally?

snreynolds: imo the hook can query the contract to check expiry since it has the keyHash

test/libraries/SettingsLib.t.sol
```diff
@@ -42,13 +42,15 @@ contract SettingsLibTest is Test {
...
+        (bool expired,) = settings.isExpired();
+        assertEq(expired, false);
     }
 
     function test_isExpired_fuzz(uint40 expiration) public {
         vm.assume(expiration > 0 && expiration < type(uint40).max);
         Settings settings = SettingsBuilder.init().fromExpiration(expiration);
         vm.warp(expiration + 1);
-        assertEq(settings.isExpired(), true);
+        (bool expired,) = settings.isExpired();
```
zhongeric, line 53: nit: check expiry here against fuzzed input

zhongeric (approved):

PR #123, opened 2025-04-09
Kleinvieh macht auch Mist

snreynolds:


foundry.toml
```diff
@@ -3,14 +3,19 @@ src = "src"
...
 ffi = true
 
 remappings = [
     "webauthn-sol/=lib/webauthn-sol/"
 ]
+additional_compiler_profiles = [{ name = "via_ir", optimizer_runs = 10000, via_ir = true }]
+
+compilation_restrictions = [
+  { paths = "src/MinimalDelegationEntry.sol", via_ir = true }
+]
```
zhongeric, line 18: so this makes it so only the entry.sol is compiled w ir, not the tests?

zhongeric (approved):

foundry.toml
snreynolds: yep!

PR #127, opened 2025-04-10
Add Base Validation, Clean Inheritance

snreynolds:


src/ERC7201.sol
```diff
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.29;
 
+import {IERC7201} from "./interfaces/IERC7201.sol";
+
 /// @title ERC-7201
 /// @notice Public getters for the ERC7201 calculated storage root, namespace, and version
-contract ERC7201 {
+contract ERC7201 is IERC7201 {
```
zhongeric, line 8: should we make this abstract?

zhongeric (approved):

src/ERC7201.sol
snreynolds: don't think it needs to be! no virtual functions

PR #129, opened 2025-04-10
fix: Readme

zhongeric:


dianakocsis:
do u also want to create a diagram for the `updateEntryPoint` flow? seems like all the other ones are covered except that one!

dianakocsis (approved):

zhongeric:
> do u also want to create a diagram for the `updateEntryPoint` flow? seems like all the other ones are covered except that one!

Ah yeah maybe I can add, I'm sure we'll udpate the README more

PR #134, opened 2025-04-10
Small fixes: tighten pragmas, licenses, unusued functions, natspec

zhongeric:


LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
snreynolds, line 1: use UNLICENSED for now

src/NonceManager.sol
```diff
@@ -1,12 +1,13 @@
...
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.29;
 
 import {INonceManager} from "./interfaces/INonceManager.sol";
 import {BaseAuthorization} from "./BaseAuthorization.sol";
 
 /// @title NonceManager
 /// @notice A contract that manages nonces to prevent replay attacks
 abstract contract NonceManager is INonceManager, BaseAuthorization {
+    /// @notice A mapping of keys to their corresponding sequence numbers
```
snreynolds, line 10: move natspec for public/external functions/storage to interface and inheritdoc

src/libraries/CalldataDecoder.sol
```diff
@@ -1,61 +1,13 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.23;
 
-import {Call} from "./CallLib.sol";
-
+/// @title CalldataDecoder
 library CalldataDecoder {
-    using CalldataDecoder for bytes;
```
snreynolds, line 7: wowww this is nice lol

src/libraries/HooksLib.sol
```diff
@@ -31,6 +40,8 @@ library HooksLib {
         return hookValidationData;
     }
 
+    /// @notice Handles the afterIsValidSignature hook
+    /// @return magicValue the ERC-1271 magic value returned by the hook
```
snreynolds, line 44: nit: would have this return say smthn like it indicates whether the signature is valid or not, and can override internal logic

src/libraries/HooksLib.sol
```diff
@@ -1,25 +1,34 @@
...
+    /// @notice Hook did not return its selector
     error InvalidHookResponse();
 
+    /// @notice Returns whether the flag is configured for the hook
     function hasPermission(IHook self, uint160 flag) internal pure returns (bool) {
         return uint160(address(self)) & flag != 0;
     }
 
+    /// @notice Handles the afterValidateUserOp hook
+    /// @return validationData encoded according to ERC-4337 spec
```
snreynolds, line 31: same add something about how it can override internal validationData

snreynolds: huh I just thought of something.. If the hook doesn't want to override the validationData it has to figure out how to construct the validationData itself, maybe if it returns some sentinel value, it should not default override?

src/NonceManager.sol
zhongeric: i just removed, i wasn't sure how to do it for storage var, and this one was not rly useful

src/libraries/CalldataDecoder.sol
zhongeric: yeah lol 🙏

src/libraries/HooksLib.sol
zhongeric: Hmm interesting, I guess it doesn't seem very difficult for the hook to return validation data? But I agree the default override behavior is a little interesting

snreynolds: ya its not necessary but just a callout

src/libraries/HooksLib.sol
```diff
@@ -53,7 +57,7 @@ library HooksLib {
     }
 
     /// @notice Handles the afterVerifySignature hook
-    /// @dev Expected to revert if the signature is invalid
+    /// @notice MUST revert if the signature is invalid
```
snreynolds, line 60: this is not really true, if the signature is invalid we wouldn't get to the hook callsite no?

snreynolds (approved):

PR #138, opened 2025-04-11
Multi call invariant tests

zhongeric:


snreynolds (commented):
ok so this isn't really multicall yes? its just testing multiple nested self calls

snreynolds (approved):

PR #140, opened 2025-04-11
Implement 7739

zhongeric:
Implement ERC7739 per the spec: https://eips.ethereum.org/EIPS/eip-7739 , which defines dynamic nested 712 encoding. 

Files:
- ERC7739.sol 

Inspired by https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/utils/cryptography/ERC7739.sol

Base contract implementing logic for verifying nested typed data sign signatures and personal signatures. Handles decoding of wrapped signatures and generating the correct hashTypedDataV4 based on the current account's domainSeparator.

- libraries/ERC7739Utils.sol

Modified from https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/utils/cryptography/ERC7739Utils.sol

Util contract implementing `decodeContentsDescr` which iterates over a string in memory to find the `contentsName`, handling both `implicit` and `explicit` modes. For more info, see https://eips.ethereum.org/EIPS/eip-7739#contentsdescription-with-implicit-and-explicit-modes. The main change I made here was to do string operations in memory instead of calldata. I'm open to calldata just wanted to do it in memory to get it working 😂 

- libraries/TypedDataSignLib.sol

712 hashing logic for typed data sign, generates dynamic type hash and name.

- libraries/PersonalSignLib.sol

Applies simple 712 hash over a hashed eth personal sign digest.


snapshots/MinimalDelegationExecuteTest.json
```diff
@@ -1,10 +1,10 @@
 {
```
snreynolds, line 1: can we add a gas snaptshot for isValidSignature callback?

src/EIP712.sol
```diff
@@ -59,13 +77,19 @@ contract EIP712 is IEIP712, IERC5267 {
     /// @notice Returns the `domainSeparator` used to create EIP-712 compliant hashes.
     /// @return The 32 bytes domain separator result.
     function domainSeparator() public view returns (bytes32) {
```
snreynolds, line 79: this is public so we shouldnt need another internal function

src/EIP712.sol
```diff
@@ -47,6 +48,23 @@ contract EIP712 is IEIP712, IERC5267 {
             bytes32 salt,
             uint256[] memory extensions
         )
+    {
+        (fields, name, version, chainId, verifyingContract, salt, extensions) = _eip712Domain();
+    }
+
+    /// @notice Internal eip712Domain implementation for use in ERC7739
+    function _eip712Domain()
```
snreynolds, line 56: why not just make public?

src/EIP712.sol
```diff
@@ -59,13 +77,19 @@ contract EIP712 is IEIP712, IERC5267 {
...
     }
 
-    /// @notice Public getter for `_hashTypedData()` to produce a replay-safe hash from the given `hash`.
-    /// @param hash The nested typed data hash as defined by EIP-712. Assumes the hash is the result of applying EIP-712 hashStruct.
-    /// @return The corresponding replay-safe hash.
+    /// @notice Public getter for `_hashTypedData()` to produce a EIP-712 hash using this account's domain separator
+    /// @dev This is meant to be used for internal verification of SignedBatchedCalls and thus does not produce signatures that are replay-safe.
+    ///      See ERC7739 for a version that is replay-safe and can be used for external verification through ERC-1271.
+    /// @param hash The nested typed data. Assumes the hash is the result of applying EIP-712 hashStruct.
     function hashTypedData(bytes32 hash) public view virtual returns (bytes32) {
```
snreynolds, line 93: yeah same here, it feels like there is no need to define an internal function and a public function that do the same thing? i think would prefer if they were all just public functions

src/EIP712.sol
```diff
@@ -59,13 +77,19 @@ contract EIP712 is IEIP712, IERC5267 {
...
+    function _domainSeparator() internal view returns (bytes32) {
         return
             keccak256(abi.encode(_DOMAIN_TYPEHASH, _cachedNameHash, _cachedVersionHash, block.chainid, address(this)));
     }
 
-    /// @notice Public getter for `_hashTypedData()` to produce a replay-safe hash from the given `hash`.
-    /// @param hash The nested typed data hash as defined by EIP-712. Assumes the hash is the result of applying EIP-712 hashStruct.
-    /// @return The corresponding replay-safe hash.
+    /// @notice Public getter for `_hashTypedData()` to produce a EIP-712 hash using this account's domain separator
+    /// @dev This is meant to be used for internal verification of SignedBatchedCalls and thus does not produce signatures that are replay-safe.
```
snreynolds, line 90: I think you could remove this since this note is more related to 1271 than it is to 712

src/EIP712.sol
zhongeric: oh I thought internal funcs have gas savings

src/ERC7739.sol
```diff
@@ -0,0 +1,94 @@
...
+import {EIP712} from "./EIP712.sol";
+import {Key, KeyLib} from "./libraries/KeyLib.sol";
+import {TypedDataSignLib} from "./libraries/TypedDataSignLib.sol";
+import {PersonalSignLib} from "./libraries/PersonalSignLib.sol";
+
+/// @title ERC7739
+/// @notice An abstract contract that implements the ERC-7739 standard
+/// @notice This contract assumes that all data verified through ERC-1271 `isValidSignature` implements the defensive nested hashing scheme defined in EIP-7739
+/// @dev See https://eips.ethereum.org/EIPS/eip-7739
+abstract contract ERC7739 is EIP712 {
```
snreynolds, line 15: High level comment.. but why is this a contract? It seems like it might just be better as a re-useable library? All of the functions defined here are private and I feel like we could re-architect it in a way where we wouldnt have to inherit from 712 and rather pass in any domain separator info

test/js-scripts/src/sign-wrapped-typed-data.ts
```diff
@@ -0,0 +1,117 @@
...
+            types: PermitSingleTypes,
+            primaryType: 'PermitSingle',
+            message: contents,
+            verifierDomain: verifierDomain,
+        })  
+
+        // signature is first 65 bytes of wrappedSignature + 0x
+        const signatureLength = 130 // 65 * 2
+        const start = 2; // skip the first 0x
+        const signature = '0x' + wrappedSignature.slice(start, start + signatureLength);
```
zhongeric, line 106: flagging this FYI:
- Viem returns a wrapped signature so we have to slice off the first 0x and 65 bytes for the actual signature to compare against `key.verify` in ffi test

src/MinimalDelegation.sol
```diff
@@ -191,10 +194,18 @@ contract MinimalDelegation is
     {
         (bytes32 keyHash, bytes memory signature, bytes memory hookData) =
             abi.decode(wrappedSignature, (bytes32, bytes, bytes));
-        bytes32 digest = _hashTypedData(data.hashWithWrappedType());
 
         Key memory key = getKey(keyHash);
-        bool isValid = key.verify(digest, signature);
+
+        // Must be branched because we do abi decoding in memory which will throw since the ecnoding schemes are different
```
zhongeric, line 200: Flag here also that decoding in memory makes the implementation a little awkaward. Comment here says more

test/js-scripts/src/sign-wrapped-typed-data.ts
```diff
@@ -74,40 +71,31 @@ async function signWrappedTypedData(): Promise<void> {
...
             chainId: 31337, // Default Anvil chain ID
+            // salt: '0x0000000000000000000000000000000000000000000000000000000000000000' as `0x${string}`,
         }
         const verifierDomain = {
             name: VERIFIER_DOMAIN_NAME,
             version: VERIFIER_DOMAIN_VERSION,
             verifyingContract: verifyingContract,
             chainId: 31337, // Default Anvil chain ID
-            salt: '0x' as `0x${string}`,
+            salt: '0x0000000000000000000000000000000000000000000000000000000000000000' as `0x${string}`,
```
snreynolds, line 83: so just doing '0x'was causing a diff hash?

test/utils/MockERC1271VerifyingContract.sol
```diff
@@ -17,64 +19,60 @@ struct PermitDetails {
...
         return _domainSeparatorV4();
     }
 
-    /// return the full contents descriptor string
-    function contentsDescr() external pure returns (string memory) {
-        return PERMIT_SINGLE_TYPE;
+    /// @notice Per ERC-7739, use the explicit mode for content descriptor strings since the top level type
+    ///         PermitSingle is alphabetically ordered after PermitDetails
+    /// @dev return the full contents descriptor string in explicit mode
+    function contentsDescrExplicit() external pure returns (string memory) {
```
snreynolds, line 52: should we test the implicit code path as well? also against ffi?

test/js-scripts/src/sign-wrapped-typed-data.ts
```diff
@@ -33,9 +29,9 @@ const PermitSingleTypes = {
     ],
     PermitDetails: [
       { name: 'token', type: 'address' },
-      { name: 'amount', type: 'uint256' },
-      { name: 'expiration', type: 'uint256' },
-      { name: 'nonce', type: 'uint256' }
+      { name: 'amount', type: 'uint160' },
```
snreynolds, line 32: 💀

test/ERC7739.t.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.23;
 
+import {console2} from "forge-std/console2.sol";
```
snreynolds, line 4: rm

src/libraries/TypedDataSignLib.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.23;
 
+import {console2} from "forge-std/console2.sol";
```
snreynolds, line 4: rm

src/libraries/PersonalSignLib.sol
```diff
@@ -0,0 +1,13 @@
...
+pragma solidity ^0.8.23;
+
+library PersonalSignLib {
+    bytes private constant PERSONAL_SIGN_TYPE = "PersonalSign(bytes prefixed)";
+    bytes32 private constant PERSONAL_SIGN_TYPEHASH = keccak256(PERSONAL_SIGN_TYPE);
+
+    /// @notice We don't care how the hash was computed for personal sign, and it does not match the typestring above
+    /// i.e.  keccak256("\x19Ethereum Signed Message:\n" || len(someMessage) || someMessage)
+    function hash(bytes32 message) internal pure returns (bytes32) {
+        return keccak256(abi.encode(PERSONAL_SIGN_TYPEHASH, message));
```
snreynolds, line 11: hm this seems wrong? the personal sign type defines the input as a bytes param and here it is already a bytes32?

test/js-scripts/src/sign-wrapped-typed-data.ts
zhongeric: no, this one was fine, it was in the app domain above, in solidity we didn't have the salt at all

test/utils/MockERC1271VerifyingContract.sol
zhongeric: yes good idea

src/libraries/PersonalSignLib.sol
zhongeric: Yep I asked viem to see what the reasoning is

src/ERC7739.sol
zhongeric: Ok I moved some funcs to the library so this base contract is mostly just `_isValidTypedDataSig` and `_isValidNestedPersonalSignature`. I think it might be a bit tricky to not inherit from 712 because we need to use the 712 domain in two different ways (domain separator for personal sign) and domainBytes for nested 712

src/ERC1271.sol
```diff
@@ -2,15 +2,29 @@
...
+    mapping(address => bool) private _erc1271CallerIsSafe;
+
     /// @dev The magic value returned by `isValidSignature()` if the signature is valid.
     bytes4 internal constant _1271_MAGIC_VALUE = 0x1626ba7e;
     /// @dev The magic value returned by `isValidSignature()` if the signature is invalid.
     bytes4 internal constant _1271_INVALID_VALUE = 0xffffffff;
 
+    /// @dev Returns whether the caller is considered safe, such
+    /// that we don't need to use the nested EIP-712 workflow as defined by ERC-7739
+    function isSafeERC1271Caller(address caller) public view virtual returns (bool) {
```
snreynolds, line 19: just make the mapping public?

src/libraries/PersonalSignLib.sol
snreynolds: Can we bring this up with Spearbit bc this still doesn't make sense to me why this is allowed?

src/MinimalDelegation.sol
```diff
@@ -183,18 +184,45 @@ contract MinimalDelegation is
...
             abi.decode(wrappedSignature, (bytes32, bytes, bytes));
-        bytes32 digest = _hashTypedData(data.hashWithWrappedType());
 
         Key memory key = getKey(keyHash);
-        bool isValid = key.verify(digest, signature);
+
+        bool isValid;
+        // Must be branched because we do abi decoding in memory which will throw since the encoding schemes are different
+        // ECDSA signatures are 65 bytes while P256 signatures are 64 bytes
+        if (signature.length == 64 || signature.length == 65) {
```
snreynolds, line 213: Im confused by this branching, couldn't the signature be 64 or 65 in length AND still be a personal sign?

src/ERC7739.sol
snreynolds: I guess the only thing I don't like about this structure is that EIP712 is something we need to support regardless of 7739...

If one day we decide 7739 is no longer needed, they are very tightly coupled in our contracts right now

snreynolds: Can we change the inheritance so that 712 is not a child of 7739?

zhongeric: Ok yeah I took it out

src/MinimalDelegation.sol
zhongeric: hmmm, my understanding is that personal sigs are only called offchain... so it should never trigger the safe 1271 caller path

zhongeric: and doesn't this codepath allow for personal sigs of length 64 | 65?

src/MinimalDelegation.sol
```diff
@@ -191,18 +193,41 @@ contract MinimalDelegation is
     }
 
     /// @inheritdoc ERC1271
+    /// @dev WrappedSignature is used for both NestedTypedDataSign and NestedPersonalSign signatures.
+    ///      If the caller implementing ERC-1271 is considered safe, we verify the signature over the data directly
+    /// TypedDataSign signatures are of the form: abi.encode(bytes32, bytes(`signature ‖ APP_DOMAIN_SEPARATOR ‖ contentsHash ‖ contentsDescr ‖ uint16(contentsDescr.length)`))
```
snreynolds, line 198:
```suggestion
    /// TypedDataSign signatures are of the form: abi.encode(bytes32, bytes(`signature ‖ APP_DOMAIN_SEPARATOR ‖ contentsHash ‖ contentsDescr ‖ uint16(contentsDescr.length)`), bytes)
```

snreynolds: I think missing the bytes hookData encoding?

snreynolds: Ah wait I see, maybe we need two natspec comments here 
1- on the outermost wrapped sig
2- on the innermost signature if it is a 7339 TypedDataSign compliant

src/MinimalDelegation.sol
```diff
@@ -191,18 +193,41 @@ contract MinimalDelegation is
...
+    /// No extra data is needed for NestedPersonalSign signatures.
     function isValidSignature(bytes32 data, bytes calldata wrappedSignature)
```
snreynolds, line 200: nit: name digest? then we can use name digest in the innermost calls to _isValidTypedDataSig / _isValidNestedPersonalSignature ?

src/MinimalDelegation.sol
```diff
@@ -191,18 +193,41 @@ contract MinimalDelegation is
...
         Key memory key = getKey(keyHash);
-        bool isValid = key.verify(digest, signature);
+
+        bool isValid;
+        if (erc1271CallerIsSafe[msg.sender]) {
+            isValid = key.verify(data, signature);
+        } else {
+            // We only support PersonalSign for ECDSA keys
+            if (signature.length == 65) {
+                isValid = _isValidNestedPersonalSignature(key, data, domainSeparator(), signature);
```
snreynolds, line 225: nit: if shortening to sig in the name do it here as well for consistency?
so either _isValidNestedPersonalSignature and _isValidTypedDataSignature 

or
_isValidNestedPersonalSig and _isValidTypedDataSig

src/ERC7739.sol
```diff
@@ -0,0 +1,66 @@
...
+    function _isValidTypedDataSig(Key memory key, bytes32 hash, bytes memory domainBytes, bytes memory wrappedSignature)
+        internal
+        view
+        returns (bool)
+    {
+        // uint16(contentsDescription.length) is not used since we do memory decoding right now
+        (bytes memory signature, bytes32 appSeparator, bytes32 contentsHash, string memory contentsDescr,) =
+            abi.decode(wrappedSignature, (bytes, bytes32, bytes32, string, uint16));
+
+        if (bytes(contentsDescr).length == 0) return false;
```
snreynolds, line 45: why does the spec require sending in the contentsDescr length?

src/ERC7739.sol
```diff
@@ -0,0 +1,66 @@
...
+
+        (string memory contentsName, string memory contentsType) = ERC7739Utils.decodeContentsDescr(contentsDescr);
+
+        if (bytes(contentsName).length == 0) return false;
```
snreynolds, line 49: do we not need to length check the type as well?

snreynolds: Could these length checks be all done inside .decodeContentsDescr() ?

src/ERC7739.sol
zhongeric: I think because it assumes you are decoding in calldata, but can double check

src/ERC7739.sol
zhongeric: Yes, I have that implemented in #146

zhongeric: Ok I changed this to check both, we don't need to check `contentsDescr` however because that's done in `decodeContentsDescr`

test/utils/TypedDataSignBuilder.sol
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /// @notice Produces an EIP-712 compatible digest by hashing a TypedDataSign struct against the app's domain separator
+    function hashTypedDataSign(
+        bytes32 contentsHash,
+        bytes memory domainBytes,
+        bytes32 appSeparator,
+        string memory contentsDescr
+    ) internal pure returns (bytes32) {
+        (string memory contentsName, string memory contentsType) = ERC7739Utils.decodeContentsDescr(contentsDescr);
```
snreynolds, line 28: Do we need to check length on contentsName and contentsType here as well?

zhongeric: I can add a revert but this is only used in tests

src/ERC7739.sol
```diff
@@ -0,0 +1,61 @@
...
+/// @title ERC7739
+/// @notice An abstract contract that implements the ERC-7739 standard
+/// @notice This contract assumes that all data verified through ERC-1271 `isValidSignature` implements the defensive nested hashing scheme defined in EIP-7739
+/// @dev See https://eips.ethereum.org/EIPS/eip-7739
+abstract contract ERC7739 {
+    using ERC7739Utils for *;
+    using KeyLib for Key;
+
+    /// @notice Verifies that the claimed contentsHash hashed with the app's separator matches the isValidSignature provided data
+    /// @dev This is a necessary check to ensure that the caller provided contentsHash is correct
```
snreynolds, line 19: Also that the caller provided appSeparator is true right?

src/MinimalDelegation.sol
```diff
@@ -179,18 +180,47 @@ contract MinimalDelegation is
     }
 
     /// @inheritdoc ERC1271
-    function isValidSignature(bytes32 data, bytes calldata wrappedSignature)
+    /// @dev WrappedSignature is used for both NestedTypedDataSign and NestedPersonalSign signatures.
+    ///      If the caller implementing ERC-1271 is considered safe, we verify the signature over the data directly
+    /// The signature is wrapped with the keyhash, signature, and any optional hook data
+    /// - If the caller is considered safe then we verify the signature over the data directly
```
snreynolds, line 186: i think this is commented twice?

src/libraries/ERC7739Utils.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: MIT
+
```
snreynolds, line 2: nit extra whitespace?

src/libraries/ERC7739Utils.sol
```diff
@@ -0,0 +1,96 @@
...
+        string memory contentsType
+    ) internal pure returns (bytes32) {
+        bytes32 typedDataSignHash = TypedDataSignLib.hash(contentsName, contentsType, contentsHash, domainBytes);
+        return MessageHashUtils.toTypedDataHash(appSeparator, typedDataSignHash);
+    }
+
+    /// @notice Parse the type name out of the ERC-7739 contents type description. Supports both the implicit and explicit modes
+    /// @dev Returns empty strings if the contentsDescr is invalid, which must be handled by the calling function
+    /// @return contentsName The type name of the contents
+    /// @return contentsType The type description of the contents
```
snreynolds, line 47: Should we comment link to OZ code?

test/MinimalDelegation.storage.t.sol
```diff
@@ -17,12 +17,13 @@ contract MinimalDelegationStorageTest is DelegationHandler {
      * 4: mapping(uint256 key => uint256 seq) nonceSequenceNumber
      * 5: mapping(address => uint256) allowance;
      */
-    uint256 private constant ENTRY_POINT_SLOT = 0;
-    uint256 private constant KEY_EXTRA_STORAGE_SLOT = 1;
-    uint256 private constant KEY_STORAGE_SLOT = 2;
-    uint256 private constant KEY_HASHES_SLOT = 3;
-    uint256 private constant NONCE_SEQUENCE_NUMBER_SLOT = 4;
-    uint256 private constant ALLOWANCE_SLOT = 5;
+    uint256 private constant ERC1271_CALLER_IS_SAFE_SLOT = 0;
```
snreynolds, line 20: tiniest nit ever but the comments above are super helpful imo so we should update them?

src/MinimalDelegation.sol
zhongeric: ah i removed

src/libraries/ERC7739Utils.sol
zhongeric: yep, I linked at top but will re-add here

snreynolds (approved):
Screenshot 2025-04-22 at 2 48 26 AM


😨 😨 😨 😨 😨 😨 😨 😨

PR #141, opened 2025-04-11
multicall

dianakocsis:


test/MinimalDelegation.execute.t.sol
```diff
@@ -628,4 +628,103 @@ contract MinimalDelegationExecuteTest is TokenHandler, HookHandler, ExecuteFixtu
...
+
+        BatchedCall memory batchedCall2 = CallUtils.initBatchedCall().withCalls(calls2).withShouldRevert(true);
+        SignedBatchedCall memory signedBatchedCall2 = CallUtils.initSignedBatchedCall().withBatchedCall(batchedCall2)
+            .withNonce(DEFAULT_NONCE + 1).withKeyHash(p256Key2.toKeyHash());
+        bytes memory signature2 = p256Key2.sign(signerAccount.hashTypedData(signedBatchedCall2.hash()));
+
+        // Create multicall data
+        bytes[] memory multicallData = new bytes[](2);
+        multicallData[0] = abi.encodeWithSelector(
+            bytes4(keccak256("execute((((address,uint256,bytes)[],bool),uint256,bytes32),bytes)")),
```
zhongeric, line 662: nit you should use a test builder library for this, we should have something in either CallUtils or otherwise

test/MinimalDelegation.execute.t.sol
```diff
@@ -628,4 +628,103 @@ contract MinimalDelegationExecuteTest is TokenHandler, HookHandler, ExecuteFixtu
...
+            signedBatchedCall2, // the SignedBatchedCall struct
+            signature2
+        );
+
+        // Execute the multicall
+        vm.prank(address(signerAccount));
+        signerAccount.multicall(multicallData);
+        vm.snapshotGasLastCall("multicall");
+    }
```
zhongeric, line 729: I'd want to see a test showing that the multicall passes on the original call's context:
- multicall with `execute(Signed)` and `execute()` by p256 key, first one succeeds but second one fails because the key is not root user
- multicall with root key can call multiple functions on address(this)

test/libraries/SignedBatchedCallLib.t.sol
```diff
@@ -30,9 +30,8 @@ contract SignedBatchedCallLibTest is Test {
             CallUtils.initSignedBatchedCall().withBatchedCall(batchedCall).withNonce(nonce).withKeyHash(keyHash);
         bytes32 actualHash = signedBatchedCall.hash();
 
-        bytes32 expectedHash = keccak256(
-            abi.encode(SignedBatchedCallLib.SIGNED_BATCHED_CALL_TYPEHASH, batchedCall.hash(), nonce, keyHash)
-        );
+        bytes32 expectedHash =
```
zhongeric, line 33: seems like one of these forge fmt is off lol

zhongeric (approved):

PR #142, opened 2025-04-11
admin keys can call execute

snreynolds:


snreynolds:
the downside is that if you wanted to add a hook on an admin key, it wouldn't get executed in direct calls to register/revoke

src/KeyManagement.sol
```diff
@@ -56,14 +58,35 @@ abstract contract KeyManagement is IKeyManagement, BaseAuthorization {
...
+        if (settings.isAdmin()) return true;
+        return false;
+    }
+
+    /// @inheritdoc BaseAuthorization
+    /// @dev This function is a wrapper around isAuthorizedAdmin(bytes32) to abstract the keyHash/key implememtation from the BaseAuthorization contract.
+    function isAuthorizedAdmin(address caller) public view override(BaseAuthorization) returns (bool) {
+        Key memory key = Key({keyType: KeyType.Secp256k1, publicKey: abi.encode(caller)});
+        return isAuthorizedAdmin(key.hash());
+    }
```
zhongeric, line 91: oh jeez I see what you mean by the abstraction being weird here...

src/KeyManagement.sol
```diff
@@ -56,14 +58,35 @@ abstract contract KeyManagement is IKeyManagement, BaseAuthorization {
...
+        if (keyHash.isRootKey()) return true;
+        Settings settings = keySettings[keyHash];
+        if (settings.isAdmin()) return true;
+        return false;
+    }
+
+    /// @inheritdoc BaseAuthorization
+    /// @dev This function is a wrapper around isAuthorizedAdmin(bytes32) to abstract the keyHash/key implememtation from the BaseAuthorization contract.
+    function isAuthorizedAdmin(address caller) public view override(BaseAuthorization) returns (bool) {
+        Key memory key = Key({keyType: KeyType.Secp256k1, publicKey: abi.encode(caller)});
```
zhongeric, line 89: Would add a note how it's fine to set keyType to k1 here because these require originating transactions

zhongeric:
> the downside is that if you wanted to add a hook on an admin key, it wouldn't get executed in direct calls to register/revoke

Could the modifier `onlyAdmin` be split up into a function that returns the `keyHash` for the current caller and one that checks it is an admin?

zhongeric (approved):

PR #147, opened 2025-04-17
audit(spearbit): #8

zhongeric:


src/libraries/SettingsLib.sol
```diff
@@ -45,6 +45,6 @@ library SettingsLib {
     function isExpired(Settings settings) internal view returns (bool _isExpired, uint40 _expiration) {
         uint40 _exp = expiration(settings);
         if (_exp == 0) return (false, 0);
-        return (_exp <= block.timestamp, _exp);
+        return (_exp < block.timestamp, _exp);
```
snreynolds, line 48: Update the natspec? Also you could add that this is to adhere to 4337 spec?

snreynolds (approved):

PR #148, opened 2025-04-17
audit(spearbit): #6

zhongeric:


src/libraries/SettingsLib.sol
```diff
@@ -5,7 +5,7 @@ import {IHook} from "../interfaces/IHook.sol";
 
 type Settings is uint256;
 
-/// The most significant 8 bits are reserved to specify if the key is an admin key or not. An admin key is allowed to self-call.
+/// The most significant 7 bits are reserved to specify if the key is an admin key or not. An admin key is allowed to self-call.
```
snreynolds, line 8: I think we can just re-write this entire documentation to specify the bytes used?

MS 6 bytes : UNUSED
Byte 25: isAdmin
Bytes 20-24: expiration
LS 20 bytes (bytes 0-19): VALIDATION_ADDRESS

PR #150, opened 2025-04-20
sb-1: add executor address

snreynolds:


src/MinimalDelegation.sol
```diff
@@ -182,6 +183,12 @@ contract MinimalDelegation is
         if (isExpired) revert IKeyManagement.KeyExpired(expiry);
     }
 
+    /// @notice Returns true if the msg.sender is the executor or if the executor is address(0)
+    /// @param executor The address of the allowed executor of the signed batched call
+    function _senderCanExecute(address executor) private view returns (bool) {
```
zhongeric, line 188: nit: maybe consider naming "senderIsExecutor" since can execute is related to spending guard related terminology?

zhongeric (commented):
Missing test for executor is checked?

test/MinimalDelegation.execute.t.sol
```diff
@@ -438,6 +438,88 @@ contract MinimalDelegationExecuteTest is TokenHandler, HookHandler, ExecuteFixtu
...
+            CallUtils.initSignedBatchedCall().withBatchedCall(batchedCall).withExecutor(executor);
+
+        bytes32 hashToSign = signerAccount.hashTypedData(signedCall.hash());
+        bytes memory signature = signerTestKey.sign(hashToSign);
+        bytes memory wrappedSignature = abi.encode(signature, EMPTY_HOOK_DATA);
+
+        // Even the owner on the account cannot execute the signed batched call.
+        vm.startPrank(address(signerAccount));
+        vm.expectRevert(BaseAuthorization.Unauthorized.selector);
+        signerAccount.execute(signedCall, wrappedSignature);
```
zhongeric, line 519: hmm, this seems a little odd to me. It seems to break the mental model that the owner is always fully permissioned on the account, is there a case where this would be desired behavior?

snreynolds: but then they would have set the executor to address(0) or to only themselves... I think its actually a feature.. If an executor address is set it is ONLY executable by them. Its weird if that is loop holed by the signer themself

zhongeric: i see, yeah i guess you could have a scenario where you'd want to entrust a relayer with a signature that only they can execute

zhongeric (approved):

PR #154, opened 2025-04-21
sb-13, sb-15: Do not early return from validateUserOp on invalid sig, and don't revert in expiry check

zhongeric:


src/MinimalDelegation.sol
```diff
@@ -128,29 +132,26 @@ contract MinimalDelegation is
...
         bool isValid = key.verify(userOpHash, signature);
 
-        // If signature verification failed, return failure immediately WITHOUT expiry as it cannot be trusted
-        if (!isValid) {
-            return SIG_VALIDATION_FAILED;
-        }
-
         Settings settings = getKeySettings(keyHash);
-        _checkExpiry(settings);
+        (, uint40 expiry) = settings.isExpired();
```
snreynolds, line 140: just use settings.expiration()

snreynolds (commented):
Actually I think reverting on expiration is desired, can we just do the early return fix?

src/MinimalDelegation.sol
```diff
@@ -123,6 +123,11 @@ contract MinimalDelegation is
     }
 
     /// @inheritdoc IAccount
+    /// @dev Only return validationData if the signature from the key associated with `keyHash` is valid over the userOpHash
+    ///      - The ERC-4337 spec requires that `validateUserOp` does not early return if the signature is invalid such that accurate gas estimation can be done
+    ///      - Thus, validationData can only be overriden by the hook if the signature is valid
```
zhongeric, line 128: need to change this natspec now

src/MinimalDelegation.sol
```diff
@@ -132,29 +137,15 @@ contract MinimalDelegation is
...
+        /// The userOpHash does not need to be made replay-safe, as the EntryPoint will always call the sender contract of the UserOperation for validation.
         Key memory key = getKey(keyHash);
         bool isValid = key.verify(userOpHash, signature);
 
-        // If signature verification failed, return failure immediately WITHOUT expiry as it cannot be trusted
-        if (!isValid) {
-            return SIG_VALIDATION_FAILED;
-        }
-
         Settings settings = getKeySettings(keyHash);
```
zhongeric, line 144: Add inline comment about this hook must revert to fail in validation

snreynolds: Do you think sufficient to put that in IHookValidation? Thats where I updated that behaviour for now

zhongeric: ah yeah i think so

zhongeric (commented):
lgtm

PR #159, opened 2025-04-22
fix: return bool for afterIsValidSignature hook

zhongeric:


src/MinimalDelegation.sol
```diff
@@ -237,15 +237,16 @@ contract MinimalDelegation is
 
         // Early return if the signature is invalid
         if (!isValid) return _1271_INVALID_VALUE;
-        result = _1271_MAGIC_VALUE;
 
         Settings settings = getKeySettings(keyHash);
         _checkExpiry(settings);
 
         IHook hook = settings.hook();
         if (hook.hasPermission(HooksLib.AFTER_IS_VALID_SIGNATURE_FLAG)) {
```
snreynolds, line 245: move permission check to hook lib

src/MinimalDelegation.sol
```diff
@@ -230,15 +225,12 @@ contract MinimalDelegation is
...
         _checkExpiry(settings);
 
-        IHook hook = settings.hook();
-        if (hook.hasPermission(HooksLib.AFTER_IS_VALID_SIGNATURE_FLAG)) {
-            // Hook can override the result
-            result = hook.handleAfterIsValidSignature(keyHash, digest, hookData);
-        }
+        settings.hook().handleAfterIsValidSignature(keyHash, digest, hookData);
+
+        return _1271_MAGIC_VALUE;
```
snreynolds, line 234: I think you don't need to name the return var on this function now

snreynolds (approved):

PR #170, opened 2025-04-24
Introducing.... Calibur

snreynolds:
Introducing Calibur, a derivation of _Excalibur_ and a nod to the mythical sword's reputation as a symbol of rightful authority and exceptional power.


test/js-scripts/src/utils/constants.ts
```diff
@@ -2,7 +2,7 @@ import {type Address} from "viem"
 
 
 // Define the domain name and version. Note chainId and verifyingContract are not constants.
-export const DOMAIN_NAME = 'Uniswap Minimal Delegation';
+export const DOMAIN_NAME = 'Uniswap Calibur';
```
zhongeric, line 5: nit: should we consider making the 712 name just `Calibur`

zhongeric (commented):
Missing 712 domain name change?

dianakocsis (approved):

snapshots/CaliburExecuteTest.json
```diff
@@ -1,7 +1,7 @@
 {
   "execute_BATCHED_CALL_SUPPORTS_OPDATA_singleCall": "95804",
   "execute_BATCHED_CALL_SUPPORTS_OPDATA_twoCalls": "132454",
-  "execute_BATCHED_CALL_opData_P256_singleCall": "121112",
+  "execute_BATCHED_CALL_opData_P256_singleCall": "121088",
```
zhongeric, line 4: why is this test changing but not the other ones? I thought that it could be because of the change in the 712 domain name but that should affect all signature ones?

zhongeric (approved):

snapshots/CaliburExecuteTest.json
snreynolds: calldata cost changes bc of the changing domain separator

PR #171, opened 2025-04-24
feat: Add dynamic salt to EIP-712 domain

zhongeric:


snreynolds (commented):
Can we add a test just to validate the use case of changing the salt to invalidate prior sigs?

snreynolds (approved):

PR #178, opened 2025-05-01
add events when storage mappings are updated

dianakocsis:


snapshots/ERC7914Test.json
```diff
@@ -1,5 +1,5 @@
 {
-  "approveNative": "45970",
+  "approveNative": "47378",
```
marktoda, line 2: hmm this is quite a bit

src/ERC7914.sol
```diff
@@ -14,6 +14,7 @@ abstract contract ERC7914 is IERC7914, BaseAuthorization {
     function approveNative(address spender, uint256 amount) external onlyThis returns (bool) {
         allowance[spender] = amount;
         emit ApproveNative(address(this), spender, amount);
+        emit AllowanceUpdated(spender, amount);
```
marktoda, line 17: do we really need this one and the `ApproveNative`? seems duplicative

src/NonceManager.sol
```diff
@@ -22,6 +22,7 @@ abstract contract NonceManager is INonceManager, BaseAuthorization {
         }
         nonceSequenceNumber[key] = targetSeq;
         emit NonceInvalidated(newNonce);
+        emit NonceSequenceNumberUpdated(key, targetSeq);
```
marktoda, line 25: seems duplicative with the above

src/ERC7914.sol
dianakocsis: yeah wasn't sure - was thinking the same thing

dianakocsis: the allowance also gets updated on line 72, so thought maybe they should be the same event?

src/NonceManager.sol
dianakocsis: same here - the nonce gets updated on line 44, so wasn't sure if it's helpful to make them the same event or not

src/ERC7914.sol
zhongeric: I don't think we need both, im fine with keeping approve native

zhongeric: Or, unless the spec defines that it must emit ApproveNative, could you just do AllowanceUpdated (since we use that in otehr places)

src/NonceManager.sol
zhongeric: What about removing `NonceInvalidated`? This func is updating the nonce to a new key + seq

src/ERC7914.sol
zhongeric: oh hmm, maybe we can just not emit AllowanceUpdated here and assume that approvals will be indexed differently than allowances?

src/NonceManager.sol
zhongeric: Could also do the same here, keeping NonceInvalidated and assuming that indexers will be able to support both

src/ERC7914.sol
dianakocsis: okay yeah so approvals will then have to be indexed by both `ApproveNative` and `AllowanceUpdated`?

src/NonceManager.sol
dianakocsis: yeah could do the same here - only thing is that `NonceInvalidated` only emits the newNonce, not the key and seq which might be annoying

dianakocsis:
moved to https://github.com/Uniswap/calibur/pull/183 to have correct label

PR #187, opened 2025-05-06
OZ-M03: ERC-7739 workflow deduction

zhongeric:


src/libraries/ERC7739Utils.sol
```diff
@@ -27,30 +25,30 @@ library ERC7739Utils {
...
         bytes memory domainBytes,
         bytes32 appSeparator,
-        string memory contentsName,
-        string memory contentsType
+        string calldata contentsDescr
     ) internal pure returns (bytes32) {
-        bytes32 typedDataSignHash = TypedDataSignLib.hash(contentsName, contentsType, contentsHash, domainBytes);
-        return MessageHashUtils.toTypedDataHash(appSeparator, typedDataSignHash);
+        (string calldata contentsName, string calldata contentsType) = decodeContentsDescr(contentsDescr);
+        if (bytes(contentsName).length == 0 || bytes(contentsType).length == 0) return bytes32(0);
```
zhongeric, line 36: This prevents reverts if the signature is not encoded correctly for typedDataSign

src/libraries/ERC7739Utils.sol
```diff
@@ -27,30 +25,30 @@ library ERC7739Utils {
...
+        );
     }
 
     /// @notice Parse the type name out of the ERC-7739 contents type description. Supports both the implicit and explicit modes
-    /// @dev Modified from https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/53f590e4f4902bee0e06e455332e3321c697ea8b/contracts/utils/cryptography/ERC7739Utils.sol
     /// @dev Returns empty strings if the contentsDescr is invalid, which must be handled by the calling function
     /// @return contentsName The type name of the contents
     /// @return contentsType The type description of the contents
-    function decodeContentsDescr(string memory contentsDescr)
+    function decodeContentsDescr(string calldata contentsDescr)
```
zhongeric, line 46: this function should be directly copied from https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/utils/cryptography/ERC7739Utils.sol#L168

src/libraries/ERC7739Utils.sol
```diff
@@ -75,16 +71,20 @@ library ERC7739Utils {
...
             }
         }
-        return ("", "");
+        // If we didn't find a valid contentsName, return empty strings
+        assembly ("memory-safe") {
+            contentsName.offset := 0
+            contentsName.length := 0
+            contentsType.offset := 0
+            contentsType.length := 0
+        }
```
zhongeric, line 87: Return empty strings (check this plz :))

src/libraries/KeyLib.sol
```diff
@@ -55,17 +59,21 @@ library KeyLib {
     /// @notice Verifies a signature from `key` over a `_hash`
     function verify(Key memory key, bytes32 _hash, bytes memory signature) internal view returns (bool isValid) {
         if (key.keyType == KeyType.Secp256k1) {
-            isValid = ECDSA.recover(_hash, signature) == abi.decode(key.publicKey, (address));
+            isValid = ECDSA.tryRecover(_hash, signature) == abi.decode(key.publicKey, (address));
```
zhongeric, line 58: using solady's `tryRecover` to not revert, but rather return address(0) if the signature is malformed. I prefer this over OZ's implementation which returns a bunch of other error data that we don't use

src/libraries/CalldataDecoder.sol
```diff
@@ -3,11 +3,101 @@ pragma solidity ^0.8.23;
...
+        internal
+        pure
+        returns (bytes calldata signature, bytes32 appSeparator, bytes32 contentsHash, string calldata contentsDescr)
+    {
+        signature = toBytes(data, 0);
+        assembly {
+            appSeparator := calldataload(add(data.offset, 0x20))
+            contentsHash := calldataload(add(data.offset, 0x40))
+        }
+        bytes calldata contentsDescrBytes = toBytes(data, 3);
```
snreynolds, line 70: I think you can just cast to a string and not declare the bytes variable.. but def test to make sure

src/MinimalDelegation.sol
```diff
@@ -140,24 +139,19 @@ contract MinimalDelegation is
...
-            isValid = _isValidTypedDataSig(key, digest, domainBytes(), signature);
-        }
+        /// Signature deduction flow as specified by ERC-7739
+        // If the signature contains enough data for a TypedDataSign, try the TypedDataSign flow
+        bool isValid = _isValidTypedDataSig(key, digest, domainBytes(), signature)
+        // If the signature is not valid as a TypedDataSign, try the NestedPersonalSign flow
+        || (
+            _isValidNestedPersonalSig(key, digest, domainSeparator(), signature)
+            // Finally, if the ERC1271 caller is considered safe, try the raw verification flow
+            || (erc1271CallerIsSafe[msg.sender] && key.verify(digest, signature))
```
snreynolds, line 153: i wonder if we could actually remove this logic if we introduce raw ecrecover for the eoa itselft

src/MinimalDelegation.sol
```diff
@@ -65,8 +65,7 @@ contract MinimalDelegation is
         _processBatch(batchedCall, keyHash);
     }
 
-    /// @inheritdoc IMinimalDelegation
-    function execute(SignedBatchedCall memory signedBatchedCall, bytes memory wrappedSignature) public payable {
+    function execute(SignedBatchedCall memory signedBatchedCall, bytes calldata wrappedSignature) public payable {
```
snreynolds, line 68: Should this now be 

```
function execute(SignedBatchedCall calldata signedBatchedCall, bytes calldata wrappedSignature) public payable
```

src/MinimalDelegation.sol
```diff
@@ -140,24 +139,19 @@ contract MinimalDelegation is
             }
         }
 
-        (bytes32 keyHash, bytes memory signature, bytes memory hookData) =
-            abi.decode(wrappedSignature, (bytes32, bytes, bytes));
+        (bytes32 keyHash, bytes calldata signature, bytes memory hookData) =
```
snreynolds, line 142: this is odd to me that we return part of this in calldata and part in memory?

src/libraries/ERC7739Utils.sol
snreynolds: And why do we need to prevent reverts here?

src/libraries/ERC7739Utils.sol
snreynolds: Yeah, makes sense to me, it should just be no data

src/libraries/KeyLib.sol
snreynolds: Are you saying you need it to not revert?

src/libraries/ERC7739Utils.sol
```diff
@@ -27,30 +25,30 @@ library ERC7739Utils {
...
-    /// @param contentsName The top level type, per EIP-712
-    /// @param contentsType The full type string of the contents, per EIP-712
+    /// @param contentsDescr The full type string of the contents, per EIP-712
     function toNestedTypedDataSignHash(
         bytes32 contentsHash,
         bytes memory domainBytes,
         bytes32 appSeparator,
-        string memory contentsName,
-        string memory contentsType
+        string calldata contentsDescr
```
snreynolds, line 33: Why do we change this?

src/libraries/ERC7739Utils.sol
snreynolds: Oh because we need it to fallback to other signature verification correct? Can we document that whenever we're returning something empty instead of reverting?

zhongeric: Yep, and can do re. documentation

src/libraries/CalldataDecoder.sol
```diff
@@ -3,11 +3,101 @@ pragma solidity ^0.8.23;
...
+    uint256 constant SLICE_ERROR_SELECTOR = 0x3b99b53d;
+
     /// @notice Removes the selector from the calldata and returns the encoded params.
     function removeSelector(bytes calldata data) internal pure returns (bytes calldata params) {
-        assembly {
+        assembly ("memory-safe") {
+            if lt(data.length, 4) {
+                mstore(0, SLICE_ERROR_SELECTOR)
+                revert(0x1c, 4)
+            }
```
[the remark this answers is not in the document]
zhongeric: Removed

src/MinimalDelegation.sol
zhongeric: Fixed!

src/libraries/ERC7739Utils.sol
zhongeric: I'm actually not sure, I might have been getting stack too deep before, but now it works. I'll revert it back and it'll be cleaner!

src/libraries/CalldataDecoder.sol
zhongeric: Ah yep I think it should work?

src/MinimalDelegation.sol
snreynolds: Oh sorry I meant to specifically call out the erc1271CallerIsSafe logic

i kind of think we should remove that..it's super jank

src/libraries/CalldataDecoder.sol
```diff
@@ -3,11 +3,93 @@ pragma solidity ^0.8.23;
...
+
+    /// @notice mask used for offsets and lengths to ensure no overflow
+    /// @dev no sane abi encoding will pass in an offset or length greater than type(uint32).max
+    ///      (note that this does deviate from standard solidity behavior and offsets/lengths will
+    ///      be interpreted as mod type(uint32).max which will only impact malicious/buggy callers)
+    uint256 constant OFFSET_OR_LENGTH_MASK = 0xffffffff;
+    uint256 constant OFFSET_OR_LENGTH_MASK_AND_WORD_ALIGN = 0xffffffe0;
+
+    /// error SliceOutOfBounds();
+    uint256 constant SLICE_ERROR_SELECTOR = 0x3b99b53d;
```
snreynolds, line 17: rm this we add it in the other PR when we do the length check in remove selector

src/MinimalDelegation.sol
```diff
@@ -140,24 +139,19 @@ contract MinimalDelegation is
...
-            isValid = key.verify(digest, signature);
-        } else if (msg.sender == address(0)) {
-            // We only support PersonalSign for offchain calls
-            isValid = _isValidNestedPersonalSig(key, digest, domainSeparator(), signature);
-        } else {
-            isValid = _isValidTypedDataSig(key, digest, domainBytes(), signature);
-        }
+        /// Signature deduction flow as specified by ERC-7739
+        // If the signature contains enough data for a TypedDataSign, try the TypedDataSign flow
+        bool isValid = _isValidTypedDataSig(key, digest, domainBytes(), signature)
```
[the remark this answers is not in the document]
snreynolds: I think the key thing is that IF there is more information in the signature needed to check the validity, we must not revert early in _isValidTypedDataSig or _isValidNestedPersonalSig

If that is true for ALL cases then I think its ok.

The reason it seems that its ok for webauthn and p256 keys to decode in memory in the key lib is because there is no extra data in the signature field, so for correctly formed signature fields that want to get to `erc1271CallerIsSafe` code, it will NOT revert on their decoding and just return false when trying to recover...

I will think about this some more tmr as I want to make sure that we catch all possible revert paths from inside _isValidTypedDataSig and _isValidNestedPersonalSig

test/CalldataDecoder.t.sol
```diff
@@ -1,27 +0,0 @@
-// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
zhongeric: its moved into a diff folder

snapshots/CaliburIsValidSignatureTest.json
```diff
@@ -1,8 +1,11 @@
 {
-  "isValidSignature_P256": "45287",
-  "isValidSignature_P256_typedData_notNested_safeERC1271Caller": "32773",
-  "isValidSignature_P256_withHook": "51296",
-  "isValidSignature_WebAuthnP256": "51973",
-  "isValidSignature_rootKey": "20822",
-  "isValidSignature_rootKey_typedData_notNested_safeERC1271Caller": "8304"
+  "isValidSignature_P256_personalSign": "35124",
+  "isValidSignature_P256_typedDataSign": "42515",
+  "isValidSignature_P256_typedData_notTypedDataSign_safeERC1271Caller": "295984",
```
zhongeric, line 4: why is this gas so high 💀

snreynolds: that seems odd..

snreynolds: I wonder if its not short circuiting?

snreynolds: Oh wait sorry in this case it should be going through each of the different signature verification paths, before it hits isSafeCaller, so it does make sense that its more expensive... its just quite a lot

snreynolds: Can we get a gas snap for a p256 personal sign signature? it seems like in general p256 verifications are more expensive and the only other benchmark we have here with a p256 key is when it short circuits in `isValidSignature_P256_withHook` so maybe we can see what the cost is for isValidSignature_P256_personalSign.. and roughly it should be inbetween these two gas snaps

snapshots/CaliburIsValidSignatureTest.json
```diff
@@ -1,8 +1,11 @@
...
+  "isValidSignature_P256_withHook": "48523",
+  "isValidSignature_WebAuthnP256_typedDataSign": "49106",
+  "isValidSignature_rootKey_personalSign": "10391",
+  "isValidSignature_rootKey_typedDataSign": "17780",
+  "isValidSignature_rootKey_typedData_notTypedDataSign_safeERC1271Caller": "16669",
+  "isValidSignature_sep256k1_typedDataSign": "38635"
```
snreynolds, line 10: why is this so much more than the root key signing over typed data sign?

snreynolds: both are k1 keys.. they should be relatively similar?

snreynolds: ah I guess the root key we don't need to load anything from storage

snapshots/CaliburIsValidSignatureTest.json
zhongeric: There is a test for that in #193 , but the huge gas cost here is definitely from the P256 library. It first tries to verify using the precompile, then if that fails (it does here since it is not a valid nestedPersonalSign), it tries to verify entirely in solidity....

snapshots/CaliburIsValidSignatureTest.json
zhongeric: Yea that would be my assumption too

src/ERC7739.sol
```diff
@@ -34,29 +36,29 @@ abstract contract ERC7739 {
...
-        (string memory contentsName, string memory contentsType) = ERC7739Utils.decodeContentsDescr(contentsDescr);
-        // These returned values will be empty if the contentsDescr is invalid
+        if (!_callerHashMatchesReconstructedHash(appSeparator, digest, contentsHash)) return false;
+
+        (string calldata contentsName, string calldata contentsType) = contentsDescr.decodeContentsDescr();
+        // For safety, ERC-7739 recommends to treat the signature as invalid if either the contentsName or contentsType are empty
         if (bytes(contentsName).length == 0 || bytes(contentsType).length == 0) return false;
 
-        if (!_callerHashMatchesReconstructedHash(appSeparator, digest, contentsHash)) return false;
+        bytes32 computed = contentsHash.toNestedTypedDataSignHash(domainBytes, appSeparator, contentsName, contentsType);
```
snreynolds, line 50: nit naming? maybe like rehashedDigest ?

src/libraries/CalldataDecoder.sol
```diff
@@ -3,11 +3,90 @@ pragma solidity ^0.8.23;
 
 /// @title CalldataDecoder
 library CalldataDecoder {
+    using CalldataDecoder for bytes;
+    /// @notice equivalent to SliceOutOfBounds.selector, stored in least-significant bits
```
snreynolds, line 7: nit: remove comment?

src/ERC7739.sol
zhongeric: renaming to `nestedDigest`

snreynolds (approved):
👏

PR #193, opened 2025-05-07
Add raw ECDSA path and remove erc1271SafeCaller

zhongeric:
# Add Raw ECDSA Signature Support for Root Key

## Summary
This PR adds support for raw 65-byte ECDSA signatures from the root key, while removing the problematic `erc1271SafeCaller` approach. The changes streamline signature validation and improve the security model by enforcing standard signature verification paths.

## Key Changes

1. **Raw ECDSA Fallback for Root Key**:
   - Added direct support for 65-byte signatures from the root key in `isValidSignature`
   - This enables simple signature verification without wrapping when using the root key

2. **Removed erc1271SafeCaller**

4. **Comprehensive Test Matrix**:
   - Completely restructured `Calibur.isValidSignature.t.sol` tests using a consistent naming pattern
   - Test function names now follow the format: `test_isValidSignature_{keyType}_{signatureType}_{flow}_{result}_gas()`
   - This creates a clear testing matrix across:
     - Key types (rootKey, secp256k1, P256, WebAuthnP256)
     - Signature types (rawSignature, wrappedSignature)
     - Validation flows (typedDataSign, nestedPersonalSign)
     - Results (isValid, invalidSigner)


snapshots/MinimalDelegationIsValidSignatureTest.json
```diff
@@ -1,10 +1,10 @@
 {
   "isValidSignature_P256": "40418",
   "isValidSignature_P256_personalSign": "14857",
-  "isValidSignature_P256_typedData_notTypedDataSign_safeERC1271Caller": "287028",
+  "isValidSignature_P256_typedData_notTypedDataSign_safeERC1271Caller": "287048",
   "isValidSignature_P256_withHook": "46426",
-  "isValidSignature_WebAuthnP256": "52203",
+  "isValidSignature_WebAuthnP256": "47009",
```
zhongeric, line 6: why does this decrease webauthn flow

snapshots/CaliburIsValidSignatureTest.json
```diff
@@ -1,11 +1,12 @@
...
-  "isValidSignature_rootKey_personalSign": "10391",
-  "isValidSignature_rootKey_typedDataSign": "17780",
-  "isValidSignature_rootKey_typedData_notTypedDataSign_safeERC1271Caller": "16669",
-  "isValidSignature_sep256k1_typedDataSign": "38635"
+  "isValidSignature_P256_wrappedSignature___typedDataSign": "42513",
+  "isValidSignature_P256_wrappedSignature___typedDataSign_withHook": "48522",
+  "isValidSignature_P256_wrappedSignature__nestedPersonalSign": "35076",
+  "isValidSignature_WebAuthnP256_wrappedSignature___typedDataSign": "49104",
+  "isValidSignature_WebAuthnP256_wrappedSignature__nestedPersonalSign": "41635",
+  "isValidSignature_rootKey_rawSignature": "4911",
```
snreynolds, line 7: 🤤

src/Calibur.sol
```diff
@@ -137,6 +137,12 @@ contract Calibur is
                 // Forces the compiler to optimize for smaller bytecode size.
                 if (uint256(digest) == ~wrappedSignature.length / 0xffff * 0x7739) return 0x77390001;
             }
+            // If the signature is 65 bytes, it is a raw ECDSA signature and MUST be verified as the root key
+            else if (wrappedSignature.length == 65) {
+                if (KeyLib.toRootKey().verify(digest, wrappedSignature)) {
```
snreynolds, line 142: if the signature is 65 bytes (or 64 re comment above) and it DOES not pass the verification against the root key I think we should early return false no?

snreynolds: And I think we cannot early return when it is valid as we need to still process the hook call in the valid case

snreynolds: Can you also leave a comment about why we can skip the hook callsite when the signature is valid against the root key? I got confused and then remembered we don't support hooks essentially on the root key path

src/interfaces/IERC1271.sol
```diff
@@ -3,14 +3,11 @@ pragma solidity ^0.8.0;
...
-    /// @notice Returns whether the caller is considered safe to skip the nested EIP-712 workflow
-    function erc1271CallerIsSafe(address caller) external view returns (bool);
-
-    /// @notice Sets whether the caller is considered safe to skip the nested EIP-712 workflow
-    function setERC1271CallerIsSafe(address caller, bool isSafe) external;
-
     /// @notice Validates the `signature` against the given `hash`.
-    /// @dev Wraps the given `hash` in a EIP-712 compliant struct along with the domain separator to be replay safe. Then validates the signature against it.
+    /// @dev Supports the following signature workflows:
+    /// - 65-byte ECDSA signatures from address(this)
```
snreynolds, line 8:
```suggestion
    /// - 65-byte or 64-byte ECDSA signatures from address(this)
```

test/Calibur.isValidSignature.t.sol
```diff
@@ -349,20 +372,108 @@ contract CaliburIsValidSignatureTest is DelegationHandler, HookHandler, ERC1271H
...
-    function test_isValidSignature_rootKey_notTypedDataSign_invalidSigner() public {
+    function test_isValidSignature_secp256k1_rawSignature_reverts_withKeyDoesNotExist() public {
+        TestKey memory key = TestKeyManager.withSeed(KeyType.Secp256k1, 0xb0b);
+        vm.prank(address(signer));
+        signerAccount.register(key.toKey());
+
+        bytes32 digest = keccak256("test");
+        bytes memory signature = key.sign(digest);
+
+        vm.expectRevert(IKeyManagement.KeyDoesNotExist.selector);
```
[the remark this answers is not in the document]
snreynolds: I think fixing this comment https://github.com/Uniswap/calibur/pull/193/files#r2081844092 would fix this test case

test/Calibur.isValidSignature.t.sol
```diff
@@ -349,20 +372,108 @@ contract CaliburIsValidSignatureTest is DelegationHandler, HookHandler, ERC1271H
...
+        signerAccount.isValidSignature(digest, signature);
+    }
+
+    /// P256 signatures are 64 bytes, so the first 32 bytes will be assumed to be the keyHash which will revert with KeyDoesNotExist()
```
snreynolds, line 420: I like the test tags! Could we add them here too? I think this one would be

    /**
     * Scenario: P256 key
     * 1. Raw signature
     * 2. Not typed data sign
     * 3. Not nested personal sign
     * = revert 
     */


?

snreynolds (commented):
some comments but close!

src/Calibur.sol
zhongeric: I did not return false if it failed here because we had erc1271safeCaller, but now that we removed it we should return false here. Good catch

test/Calibur.isValidSignature.t.sol
zhongeric: With your suggestion above to return false early if it is 64 or 65 but not from the root key this would be:
```
    /**
     * Scenario: P256 key
     * 1. 64 or 65-byte signature not from the root key
     * = invalid signer
     */
```

snreynolds: yep that makes sense

src/libraries/WrappedSignatureLib.sol
```diff
@@ -0,0 +1,64 @@
...
+    using CalldataDecoder for bytes;
+
+    /// @notice Returns true if the signature is empty
+    /// @dev For use in the ERC-7739 sentinel value check
+    function isEmpty(bytes calldata data) internal pure returns (bool) {
+        return data.length == 0;
+    }
+
+    /// @notice Returns true if the signature does not contain any extra data (keyHash or hookData)
+    /// @dev Technically it will return false for WebAuthn signatures but they can't be verified without a keyHash
```
snreynolds, line 18: This comments kinda confusing.. Can we maybe name this function like "isRawSignature" and just say returns true for standard ecdsa length signatures

and then @dev will also return true for standard p256 signatures however those MUST be wrapped with extra information in the verify sig flow

zhongeric: Agree, thanks!

snreynolds (approved):
lgtm

PR #195, opened 2025-05-07
OZ N-02 State Variable Visibility Not Explicitly Declared

zhongeric:


src/KeyManagement.sol
```diff
@@ -13,9 +13,9 @@ abstract contract KeyManagement is IKeyManagement, BaseAuthorization {
     using KeyLib for *;
     using SettingsLib for Settings;
 
-    EnumerableSetLib.Bytes32Set keyHashes;
-    mapping(bytes32 keyHash => bytes encodedKey) keyStorage;
-    mapping(bytes32 keyHash => Settings settings) keySettings;
+    EnumerableSetLib.Bytes32Set public keyHashes;
+    mapping(bytes32 keyHash => bytes encodedKey) public keyStorage;
+    mapping(bytes32 keyHash => Settings settings) public keySettings;
```
snreynolds, line 18: I wonder if keySettings should be private? 

We kind of want people to use getKeySettings because keySettings with the root key hash has custom logic

snreynolds: maybe actually same for keyStorage mapping

snreynolds (commented):
2 comments, and then this needs merge conf fixes toooo

src/KeyManagement.sol
zhongeric: oh good point

snreynolds (approved):

PR #204, opened 2025-05-08
fix: M02 Pass relevant information to hooks

snreynolds:


zhongeric (approved):

zhongeric (approved):

zhongeric (approved):
Will track that we should add an e2e for the hook flow

PR #208, opened 2025-05-09
OZ-L04 support prehashing for digests verified by P256 keys

zhongeric:


src/Calibur.sol
```diff
@@ -87,7 +87,7 @@ contract Calibur is
     function executeUserOp(PackedUserOperation calldata userOp, bytes32) external onlyEntryPoint {
         // Parse the keyHash from the signature. This is the keyHash that has been pre-validated as the correct signer over the UserOp data
         // and must be used to check further on-chain permissions over the call execution.
-        (bytes32 keyHash,,) = abi.decode(userOp.signature, (bytes32, bytes, bytes));
+        (bytes32 keyHash,,) = userOp.signature.decodeWithKeyHashAndHookData();
```
zhongeric, line 90: This needed to be changed in #187 where we decode signatures in calldata now. It was possible to send in a wrapped signature with:
`keyHash || signature` but no hookData which would not revert in `validateUserOp` but would revert here when you did the in memory decoding

test/utils/TestKeyManager.sol
```diff
@@ -73,7 +73,7 @@ library TestKeyManager {
     function sign(TestKey memory key, bytes32 hash) internal pure returns (bytes memory) {
         if (key.keyType == KeyType.P256) {
             (bytes32 r, bytes32 s) = vm.signP256(key.privateKey, hash);
-            return abi.encodePacked(r, s);
+            return abi.encode(r, s, false);
```
zhongeric, line 76: If we wanted to move the signature decoding into calldata too (we would have to port over webauthn decoding), we could do packed here

src/Calibur.sol
snreynolds: This doesn't NEED to be decoded in calldata afaik.. the reason we needed to decode in 187 is to prevent short circuiting but this is like a top level function call from the entrypoint

if it reverts here it just means it was malformed calldata

src/libraries/KeyLib.sol
```diff
@@ -1,9 +1,12 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.23;
 
+import {console2} from "forge-std/console2.sol";
```
snreynolds, line 4: rm

test/Calibur.isValidSignature.t.sol
```diff
@@ -421,13 +421,15 @@ contract CaliburIsValidSignatureTest is DelegationHandler, HookHandler, ERC1271H
      * 1. 64 or 65-byte signature not from the root key
      * = invalid signer
      */
+    /// @dev We generate a signature without the `preHash` flag encoded so it passes the length check for 64 || 65 bytes
```
snreynolds, line 424: rm comment

src/Calibur.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.23;
 
+import {console2} from "forge-std/console2.sol";
```
snreynolds, line 4: rm

src/libraries/WrappedSignatureLib.sol
```diff
@@ -8,6 +8,9 @@ import {CalldataDecoder} from "./CalldataDecoder.sol";
 library WrappedSignatureLib {
     using CalldataDecoder for bytes;
 
+    /// error SliceOutOfBounds();
+    uint256 constant SLICE_ERROR_SELECTOR = 0x3b99b53d;
```
snreynolds, line 12: also rm

src/Calibur.sol
snreynolds: We are changing this behaviour slightly with https://github.com/Uniswap/calibur/pull/218

src/libraries/KeyLib.sol
```diff
@@ -52,15 +56,14 @@ library KeyLib {
...
-    function verify(Key memory key, bytes32 _hash, bytes memory signature) internal view returns (bool isValid) {
+    /// @dev Signatures from P256 are expected to be over the `sha256` hash of `_hash`
+    function verify(Key memory key, bytes32 _hash, bytes calldata signature) internal view returns (bool isValid) {
         if (key.keyType == KeyType.Secp256k1) {
             isValid = ECDSA.tryRecover(_hash, signature) == abi.decode(key.publicKey, (address));
         } else if (key.keyType == KeyType.P256) {
-            // Extract x,y from the public key
             (bytes32 x, bytes32 y) = abi.decode(key.publicKey, (bytes32, bytes32));
-            // Split signature into r and s values.
             (bytes32 r, bytes32 s) = abi.decode(signature, (bytes32, bytes32));
```
snreynolds, line 65: I think you can leave the comments, maybe accidentally deleted?

snreynolds (approved):

Issue #188, opened 2025-05-06
Fix: Do not restrict direct calls to execute by admins


snreynolds:
Any non-expired, registered key should be able to directly call `execute`.

snreynolds:
Note that this means the key must be a SCEP-K1 key

PR #214, opened 2025-05-09
fix: Allow any registered,unexpired key to direct call

snreynolds:
fixes https://github.com/Uniswap/calibur/issues/188


test/utils/FunctionCallGenerator.sol
```diff
@@ -129,7 +130,11 @@ abstract contract FunctionCallGenerator is InvariantFixtures {
             if (!isRegistered) {
                 revertData = _wrapCallFailedRevertData(IKeyManagement.KeyDoesNotExist.selector);
             }
-            return _revokeCall(keyHash, revertData);
+            // Cannot revoke the rootKeyHash since it cannot be registered 
+            else if (currentKeyHash == KeyLib.ROOT_KEY_HASH) {
+                revertData = _wrapCallFailedRevertData(IKeyManagement.KeyDoesNotExist.selector);
+            }
```
zhongeric, line 136: The reason why we have to add this now is that the behavior introduced in this PR above (108-109) special case the root key hash, returning bytes32(0) when the selected key to revoke is the root key.

This is different than the existing behavior which just called `testKey.toKeyHash()`, using the underlying `KeyLib.hash()` function which does not special case the root key hash. Instead of returning bytes32(0) it would incorrectly calculate the hash based off of the root EOA's public key, which would _not_ be registered on the account and thus `isRegistered` would be false, so the revert would be caught above in lines 130-132

snreynolds: Hm I see, but I think we can just do

if (!isRegistered || currentKeyHash == KeyLib.ROOT_KEY_HASH)

zhongeric: Ah yeah sg

zhongeric (approved):

PR #218, opened 2025-05-12
revert on length mismatch

snreynolds:


test/libraries/WrappedSignatureLib.t.sol
```diff
@@ -60,4 +60,57 @@ contract WrappedSignatureLibTest is Test {
...
+
+    function test_decodeSignatureWithHookData() public view {
+        bytes memory data = abi.encode(bytes(""), bytes(""));
+        (bytes memory _arg1, bytes memory _arg2) = decoder.decodeWithHookData(data);
+        assertEq(_arg1, bytes(""));
+        assertEq(_arg2, bytes(""));
+    }
+
+    function test_decodeSignatureWithHookData_incorrectlyEncodedSignature_reverts() public {
+        bytes memory data = abi.encode(bytes32(keccak256("test")), bytes(""));
```
zhongeric, line 72: this is expected to error because the first word, which should be the length, is some bytes32 value when the signature is empty right?

test/libraries/WrappedSignatureLib.t.sol
```diff
@@ -60,4 +60,57 @@ contract WrappedSignatureLibTest is Test {
...
+        vm.expectRevert();
+        decoder.decodeWithHookData(data);
+    }
+
+    function test_decodeSignatureWithHookData_incorrectlyEncodedHookData_reverts() public {
+        bytes memory data = abi.encode(bytes(""));
```
zhongeric, line 78: This is `abi.encode(bytes(""))`
```
➜ abi.encode(bytes(""))
Type: dynamic bytes
├ Hex (Memory):
├─ Length ([0x00:0x20]): 0x0000000000000000000000000000000000000000000000000000000000000040
├─ Contents ([0x20:..]): 0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000
```

src/libraries/CalldataDecoder.sol
```diff
@@ -50,4 +52,46 @@ library CalldataDecoder {
             }
         }
     }
+
+    /// @notice Decode the `_arg`-th element in `_bytes` as `bytes`
+    /// @dev Reverts if the length is not what the encoding is expecting
+    // @param _bytes The input bytes string to extract a bytes string from
+    // @param _arg The index of the argument to extract
+    function toSafeBytes(bytes calldata _bytes, uint256 _arg) internal pure returns (bytes calldata res) {
```
hensha256, line 60: nit but i would rename `toBytesSafe` or `safeToBytes` this name confused me as to what it was going to do

zhongeric (approved):

PR #220, opened 2025-05-14
feedback: small fixes

zhongeric:


src/EIP712.sol
```diff
@@ -8,7 +8,8 @@ import {BaseAuthorization} from "./BaseAuthorization.sol";
 import {PrefixedSaltLib} from "./libraries/PrefixedSaltLib.sol";
 
 /// @title EIP712
-/// @dev This contract does not cache the domain separator and calculates it on the fly since it will change when delegated to.
+/// @dev This contract does not cache the domain separator and calculates it on the fly
```
dianakocsis, line 11: "...and instead calculates it on the fly"?

src/EIP712.sol
```diff
@@ -95,9 +96,11 @@ contract EIP712 is IEIP712, IERC5267, BaseAuthorization {
 
     /// @inheritdoc IEIP712
     function updateSalt(uint96 prefix) external onlyThis {
+        if (prefix != _saltPrefix) {
+            // per EIP-5267, emit an event to notify that the domain separator has changed
+            emit EIP712DomainChanged();
+        }
```
snreynolds, line 102: We can just do:

```
        if (prefix != _saltPrefix) {
            _saltPrefix = prefix;
            // per EIP-5267, emit an event to notify that the domain separator has changed
            emit EIP712DomainChanged();
        }
```

snreynolds: we don't need to update _saltPrefix if the exact same value is passed in

snreynolds (commented):
Confirming that this PR includes the following changes:
- The value of [fields](https://github.com/Uniswap/calibur/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/EIP712.sol#L61) should be 1f and not 0f. ✅ 

- [MASK_160_BITS](https://github.com/Uniswap/calibur/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/libraries/PrefixedSaltLib.sol#L8) is unused. ✅ 

- EIP712DomainChanged event [could be emitted](https://github.com/Uniswap/calibur/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/EIP712.sol#L100) when the salt has not changed. 🟡 --> I have one suggestion here

- view keyword [missing](https://github.com/Uniswap/calibur/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/interfaces/IERC7914.sol#L25-L28) from nativeAllowance and transientNativeAllowance in IERC7914. ✅ 

- WrappedSignatureLib is [unnecessary](https://github.com/Uniswap/calibur/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/libraries/KeyLib.sol#L25) in KeyLib. ✅ 

- 'AFTER_VALIDATE_USER_OP_FLAG' and 'BEFORE_EXECUTE_FLAG' should be afterValidateUserOp and beforeExecute respectively in [this comment](https://github.com/Uniswap/calibur/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/libraries/HooksLib.sol#L12) in HooksLib . ✅ 

- LibString import in ERC7739Utils is unnecessary. ✅ 

- Adding [domain change](https://github.com/Uniswap/minimal-delegation/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/EIP712.sol#L11) is possible when the salt is updated. ✅ 

- Adding that the salt prefix can be updated by [the owner](https://github.com/Uniswap/minimal-delegation/blob/d9b1dacc3a818eaf3d22534aa38ba72b2c56b6f1/src/EIP712.sol#L24) or admin. ✅

snreynolds (approved):

PR #224, opened 2025-05-16
natspec for root key signatures

snreynolds:


src/interfaces/IERC1271.sol
```diff
@@ -8,6 +8,11 @@ interface IERC1271 {
     /// - 64 or 65-byte ECDSA signatures from address(this)
     /// - Nested typed data signatures as defined by ERC-7739
     /// - Nested personal signatures as defined by ERC-7739
+    /// @dev A wrappedSignature contains a keyHash, signature, and any optional hook data
+    ///      `signature` can contain extra fields used for webauthn verification or ERC7739 nested typed data verification
+    /// @dev An unwrapped signature is only valid for the root key. However, the root key can also sign a 7739 rehashed signature.
+    /// It is possible for an unwrapped signature from the root key to be replayed IF the root key is registered on another wallet and that wallet
```
zhongeric, line 14: nit: replayed on `another wallet`, not this one, right?

zhongeric (approved):

src/interfaces/IERC1271.sol
snreynolds: "..IF the root key is registered on multiple wallets and more than one of the wallets does not enforce defensive rehashing"

it is true that if the root key is registered on another wallet, signs a message intended for that wallet (which doesn't do rehashing) it could be submitted to this wallet and be successful

zhongeric (approved):

PR #227, opened 2025-05-22
Permit2 with Native

codyborn:
Leverages ERC7914 + ERC20ETH to create a Permit2 signature that can be used with native ETH.


test/interfaces/IAllowanceTransfer.sol
```diff
@@ -0,0 +1,165 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+import {IEIP712} from "./IEIP712.sol";
+
+/// @title AllowanceTransfer
+/// @notice Handles ERC20 token permissions through signature based allowance setting and ERC20 token transfers by checking allowed amounts
+/// @dev Requires user's token approval on the Permit2 contract
+interface IAllowanceTransfer is IEIP712 {
```
zhongeric, line 9: is this not importable from permit2 dep?

test/interfaces/IDAIPermit.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+interface IDAIPermit {
```
zhongeric, line 4: where is this used?

test/utils/Permit2Utils.sol
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {ISignatureTransfer} from "../interfaces/ISignatureTransfer.sol";
+import {Vm} from "forge-std/Vm.sol";
+import {IPermit2} from "../interfaces/IPermit2.sol";
+
+library Permit2Utils {
```
zhongeric, line 8: Nice! Would love to have this work well with our existing testing utils like `TestKeyManager`, etc. since we have different ways of signing depending on the keyType & if it is the root key or an external signer. For example, external signers will have to sign over 7739 wrapped types. I can add it in if you're ok with it!

test/interfaces/IAllowanceTransfer.sol
codyborn: Unfortunately Permit2 has solidity pinned to `=0.8.17` preventing forge from building it together with the rest of calibur (`^0.8.23`). I had to go with a similar approach to UniswapX where we copy in these interfaces and have a custom Permit2 deploy function.

test/interfaces/IDAIPermit.sol
codyborn: good catch, it's not

test/utils/Permit2Utils.sol
codyborn: Updated! LMK if that's what you had in mind.

test/interfaces/IPermit2.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
snreynolds, line 2: Install permit2 as a dependency and import!

test/interfaces/ISignatureTransfer.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+import {IEIP712} from "./IEIP712.sol";
+
+/// @title SignatureTransfer
+/// @notice Handles ERC20 token transfers through signature based actions
+/// @dev Requires user's token approval on the Permit2 contract
+interface ISignatureTransfer is IEIP712 {
```
snreynolds, line 9: same this should be importable

test/interfaces/IERC1271.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+interface IERC1271 {
```
snreynolds, line 4: where is this being used?

test/interfaces/IAllowanceTransfer.sol
snreynolds: but the interfaces are not pinned

snreynolds: as long as you don't import Permit2.sol I don't think it should be an issue, I just tried it here https://github.com/Uniswap/calibur/pull/229

codyborn: Awesome, thanks @snreynolds

test/utils/TestKeyManager.sol
```diff
@@ -94,6 +96,51 @@ library TestKeyManager {
...
+    /// @param spender The spender address
+    function signPermitTransfer(
+        TestKey memory key,
+        ISignatureTransfer.PermitTransferFrom memory permit,
+        bytes32 typehash,
+        bytes32 witness,
+        bytes32 domainSeparator,
+        address spender
+    ) internal view returns (bytes memory) {
+        if (typehash == bytes32(0)) {
```
snreynolds, line 114: this is a super nit since I know this is just a test file, but I think this logic could be simplified slightly by just casing on typehash == bytes32(0) bc definitionally it seems that if the typehash is 0 there must be no witness

so I think you could do:
```
bytes32 tokenPermissions = keccak256(abi.encode(Permit2Utils.TOKEN_PERMISSIONS_TYPEHASH, permit.permitted));
bytes32 msgHash;
if (typehash == bytes32(0)) {
// there is no witness specific typehash defined, so there is also no witness 
msgHash = keccak256(
            abi.encodePacked(
                "\x19\x01",
                domainSeparator,
                keccak256(
                    abi.encode(
                        Permit2Utils.PERMIT_TRANSFER_FROM_TYPEHASH, tokenPermissions, spender, permit.nonce, permit.deadline
                    )
                )
            )
        );
} else {

msgHash = msgHash = keccak256(
                abi.encodePacked(
                    "\x19\x01",
                    domainSeparator,
                    keccak256(abi.encode(typehash, tokenPermissions, spender, permit.nonce, permit.deadline, witness))
                )
            );

}

return sign(key, msgHash);
```

And now you don't hash twice for msgHash in the case that there is a witness!

test/utils/TestKeyManager.sol
```diff
@@ -94,6 +96,50 @@ library TestKeyManager {
...
 
+    /// @dev Signs a permit2 transfer message
+    /// @param key The key to sign with
+    /// @param permit The permit data to sign
+    /// @param typehash The typehash to use (0 for default PERMIT_TRANSFER_FROM_TYPEHASH)
+    /// @param witness Optional witness data to include in the signature
+    /// @param domainSeparator The domain separator for the permit2 contract
+    /// @param spender The spender address
+    function signPermitTransfer(
+        TestKey memory key,
```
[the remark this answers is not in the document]
codyborn: Just added support in TestKeyManager for external signers + some tests!

test/ERC7914.t.sol
```diff
@@ -359,4 +362,146 @@ contract ERC7914Test is DelegationHandler {
...
+        assertEq(bob.balance, spendAmount);
+        assertEq(address(signerAccount).balance, totalAmount - spendAmount);
+    }
+
+    // Test that a permit2 signature from a 7739 signer can be used to
+    // transfer native ETH using the ERC20-eth contract (no witness)
+    function test_permit2SignatureTransferNativeWith7739() public {
+        // Deploy ERC20ETH and Permit2
+        ERC20ETH erc20Eth = new ERC20ETH();
+        IPermit2 permit2 = IPermit2(Permit2Utils.deployPermit2());
```
zhongeric, line 443: When you deploy permit2 like this from bytecode what is the chainId? It is immutable (so in the deployment bytecode), which affects the domain separator.

test/utils/TestKeyManager.sol
```diff
@@ -140,6 +143,62 @@ library TestKeyManager {
...
+            permit2DomainSeparator,
+            witness == bytes32(0) ? Permit2Utils.PERMIT_TRANSFER_FROM_NAME : Permit2Utils.PERMIT_WITNESS_TRANSFER_FROM_NAME,
+            contentType
+        );
+        
+        // Sign the nested digest
+        bytes memory signature = sign(key, nestedDigest);
+        
+        // Build the ERC7739 wrapped signature
+        return TypedDataSignBuilder.buildTypedDataSignSignature(
```
zhongeric, line 194: Ideally we could follow the same format as the rest of the tests in the repo which touch the 7739 path (see Calibur.isValidSignature.t.sol):

```
        (bytes32 appDomainSeparator, string memory contentsDescr, bytes32 contentsHash) = getERC1271Fixtures();
        bytes memory signature = p256Key.sign(TEST_TYPED_DATA_SIGN_DIGEST);
        bytes memory typedDataSignSignature =
            TypedDataSignBuilder.buildTypedDataSignSignature(signature, appDomainSeparator, contentsHash, contentsDescr);
        bytes memory wrappedSignature = abi.encode(p256Key.toKeyHash(), typedDataSignSignature, EMPTY_HOOK_DATA);

        // Digest is what is calculated by the ERC1271 contract which hashes its domain separator to the contents hash
        bytes32 digest = mockERC1271VerifyingContract.hashTypedDataV4(contentsHash);
        vm.prank(address(mockERC1271VerifyingContract));
        bytes4 result = signerAccount.isValidSignature(digest, wrappedSignature);
```

Where the Permit2Utils contract can return the `contentsHash`, `contentsDesc`, and `appDomainSeparator`, and we can generate the 7739 digest, then sign it + wrap it like any other test. I feel like it's more readable than having it in this test file!

test/ERC7914.t.sol
codyborn: It looks like Permit2 (via [EIP712](https://github.com/Uniswap/permit2/blob/cc56ad0f3439c502c246fc5cfcc3db92bb8b7219/src/EIP712.sol#L20)) sets the chainId based on block.chainId so it should be dependent upon where it's deployed. LMK if I misunderstood your question.

zhongeric: Right but I think when you deploy it like this via bytecode, since immutables are part of the bytecode, it must include a chainId already. I'm just not sure which chain id haha

codyborn: Because the value it's immutable (not constant) and is assigned within the constructor, it pulls it at deployment time (not compile time).
https://docs.soliditylang.org/en/latest/contracts.html#constant-and-immutable-state-variables
> for constant variables, the value has to be fixed at compile-time, while for immutable, it can still be assigned at construction time.

test/utils/TestKeyManager.sol
```diff
@@ -140,6 +143,62 @@ library TestKeyManager {
...
+        if (witness == bytes32(0)) {
+            // Non-witness case: use PERMIT_TRANSFER_FROM_TYPEHASH
+            contentsHash = keccak256(abi.encode(Permit2Utils.PERMIT_TRANSFER_FROM_TYPEHASH, tokenPermissions, spender, permit.nonce, permit.deadline));
+            contentType = Permit2Utils.PERMIT_TRANSFER_FROM_CONTENT_TYPE;
+            contentsDescr = string(abi.encodePacked(contentType, Permit2Utils.PERMIT_TRANSFER_FROM_NAME));
+        } else {
+            // Witness case: use provided typehash
+            contentsHash = keccak256(abi.encode(typehash, tokenPermissions, spender, permit.nonce, permit.deadline, witness));
+            contentType = Permit2Utils.FULL_EXAMPLE_WITNESS_TYPE;
+            contentsDescr = string(abi.encodePacked(contentType, Permit2Utils.PERMIT_WITNESS_TRANSFER_FROM_NAME));
```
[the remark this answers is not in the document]
codyborn: In this case, `FULL_EXAMPLE_WITNESS_TYPE` is a combination of `PermitWitnessTransferFrom(TokenPermissions permitted,address spender,uint256 nonce,uint256 deadline,` + `MockWitness witness)MockWitness(uint256 value,address person,bool test)TokenPermissions(address token,uint256 amount)`

test/ERC7739.t.sol
```diff
@@ -112,4 +124,205 @@ contract ERC7739Test is DelegationHandler, TokenHandler, ERC1271Handler, FFISign
...
+        // This should succeed without reverting, proving the signature is valid
+        permit2.permit(address(signerAccount), permitSingle, wrappedSignature);
+        
+        // Verify the allowance was actually set
+        (uint160 allowance, uint48 expiration, uint48 nonce) = permit2.allowance(address(signerAccount), address(tokenA), address(this));
+        assertEq(allowance, allowanceAmount);
+        assertEq(expiration, uint48(block.timestamp + 1 hours));
+        assertEq(nonce, 1); // nonce should be incremented after permit
+
+        // Now verify the spender can actually pull the funds
```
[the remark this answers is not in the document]
codyborn: Done. I'll keep the test_signTypedSignData_permitSingleTransfer_transferNative as is, since I want to make sure native token is the thing that is sent.

zhongeric (approved):

barthezkimani:
Withdraw

PR #233, opened 2025-06-17
7914 detection

codyborn:
A util contract for checking if a 7702 wallet supports 7914 or not.

Checks
- the address is a contract
- The contract is a 7702 delegate
- if the delegate is calibur, return true
- else, uses staticcall to see if the contract supports `IERC7914.approveNative()`


zhongeric (commented):
I like the approach! Consider checking `transferFromNative` instead of `approveNative`? I think it is possible that a contract doesn't track any allowance state internally. Also, maybe we should add a step in between checking against calibur address and doing the function selector fuzzing checking if they implemented ERC165: https://github.com/ethereum/ERCs/blob/8380220418521ff1995445cff5ca1d0e496a3d2d/ERCS/erc-7914.md#erc-165-interface

